PR #1, opened 2015-10-19
Finish the service

djmitche:
This makes a bunch of changes.  The stuff that's most interesting:
 * uses `cron-parser` for schedules
 * tasks are created using a hook-specific role, which may have more permissions than the user creating/updating the hook
 * removes triggering support temporarily (since payloads weren't implemented yet)
 * lots of new tests and corresponding refactors and bugfixes


djmitche:
There are lots of to-do items after this (https://public.etherpad-mozilla.org/p/tc-hooks-tasklist) but I'd like to get something deployed so we can start experimenting with it.

hooks/taskcreator.js
```diff
@@ -0,0 +1,93 @@
...
+    assert(options, "options must be given");
+    assert(options.credentials instanceof Object,
+        "Expected credentials");
+    this.credentials = options.credentials;
+  }
+
+  taskForHook(hook) {
+    let task = _.cloneDeep(hook.task);
+    task.created = new Date().toJSON();
+    task.deadline = taskcluster.fromNow(hook.deadline).toJSON();
```
jonasfj: there is an actual `taskcluster.fromNowJSON()` method too...
Oh, and `taskcluster.fromNowJSON() === new Date().toJSON()`

hooks/taskcreator.js
```diff
@@ -0,0 +1,93 @@
...
+   * }
+   * */
+  constructor(options) {
+    assert(options, "options must be given");
+    assert(options.credentials instanceof Object,
+        "Expected credentials");
+    this.credentials = options.credentials;
+  }
+
+  taskForHook(hook) {
```
jonasfj: This needs to take a date object called `created`. You can give it a default value of `new Date()`

And all timestamps has to relative to `created`, example:
```js
taskForHook(hook, created = new Date()) {
  assert(created instanceof Date);
  task.created = created.toJSON();
  task.deadline = taskcluster.fromNowJSON(hook.deadline, created);
  ...
}
```

Then use the nextSchedule timestamp as `created` when creating for scheduled task.. Otherwise you're task definition won't be the same, and the idempotent operation will fail if you crash before saving state...

hooks/taskcreator.js
```diff
@@ -0,0 +1,93 @@
...
+    let task = _.cloneDeep(hook.task);
+    task.created = new Date().toJSON();
+    task.deadline = taskcluster.fromNow(hook.deadline).toJSON();
+    if (hook.expires) {
+      task.expires = taskcluster.fromNow(hook.expires).toJSON();
+    }
+    return Promise.resolve(task);
```
jonasfj: Why return a promise when the method is synchronous?

Note, there is nothing wrong with calling `await "my-value"`, it's just a waste.. but no need to return promises...

hooks/taskcreator.js
```diff
@@ -0,0 +1,93 @@
...
+    let queue = new taskcluster.Queue({
+      credentials: this.credentials,
+      authorizedScopes: [role]
+    });
+
+    // TODO: payload is ignored right now
+
+    debug('firing hook %s/%s to create taskId: %s',
+        hook.hookGroupId, hook.hookId, taskId);
+    return await queue.createTask(taskId, await this.taskForHook(hook));
```
jonasfj: No need to await `await this.taskForHook(hook)` just do `this.taskForHook(hook)`... well add the created we talked about... it can probably live in options :)

hooks/taskcreator.js
```diff
@@ -0,0 +1,93 @@
...
+  * Fire the given hook, using the given payload (interpolating it into the task
+  * definition).  If options.taskId is set, it will be used as the taskId;
+  * otherwise a new taskId will be created.
+  */
+  async fire(hook, payload, options) {
+    options = options || {};
+    var taskId = options.taskId;
+    if (!taskId) {
+      taskId = slugid.v4();
+    }
```
jonasfj: It's a matter of taste I prefer to do these with:
```js
options = _.defaults({}, options, {
  taskId: taskcluster.slugid(), // use this over slugid module
  created: new Date(),
});
```
It feels more declarative... but it's no worries.. just a random note...

hooks/taskcreator.js
```diff
@@ -0,0 +1,93 @@
...
+
+    // TODO: payload is ignored right now
+
+    debug('firing hook %s/%s to create taskId: %s',
+        hook.hookGroupId, hook.hookId, taskId);
+    return await queue.createTask(taskId, await this.taskForHook(hook));
+  };
+}
+
+module.exports.TaskCreator = TaskCreator;
```
jonasfj: this is just:
`exports.TaskCreator = TaskCreator` same below...

hooks/nextdate.js
```diff
@@ -0,0 +1,25 @@
...
+        utc: true
+      });
+    let n = interval.next();
+    if (typeof next === 'undefined' || n < next) {
+        next = n;
+    }
+  });
+
+  // always return a date, even if it's 1970
+  return next || new Date(0);
```
jonasfj: maybe MAX_INT is better or something... it just seems that this doesn't fly well with:
`nextScheduledDate:  base.Entity.op.lessThan(new Date())` later...

I know setting `MAX_INT` would be a hack...

hooks/scheduler.js
```diff
@@ -49,7 +51,7 @@ class Scheduler {
     this.stopping = false;
 
     // Create a promise that we're done looping
-    this.done = poll().catch((err) => {
+    this.done = this.loopUntilStopped().catch((err) => {
       debug("Error: %s, as JSON: %j", err, err, err.stack);
       throw err;
```
jonasfj: This needs to `this.emit('error', err);` and we have to subclass eventEmitter, `class Scheduler extends events.EventEmitter`.

jonasfj: Node has a special trick for unhandled `error` events, they crash the app... Which is really what we want in this case.

hooks/scheduler.js
```diff
@@ -63,24 +65,27 @@ class Scheduler {
...
   async poll() {
-    while(!this.stopping) {
-      // Get all hooks that have a scheduled date that is earlier than now
-      var hooks = await this.Hook.scan({
-        nextScheduledDate:  base.Entity.Op.lessThan(new Date())
-      }, {});
+    // Get all hooks that have a scheduled date that is earlier than now
+    var hooks = await this.Hook.scan({
+      nextScheduledDate:  base.Entity.op.lessThan(new Date())
+    }, {});
```
jonasfj: This needs to handle paging... azure may page if crossing partition boundaries, or being slow...
It's easy to do this with, handler: https://github.com/taskcluster/azure-entities/blob/c2a386a5a4d33d791ce8ef27c7a03ba4962d868b/src/entity.js#L1156-L1159

Example:
```js
var hooks = await this.Hook.scan({
  nextScheduledDate:  base.Entity.op.lessThan(new Date())
}, {
  limit: 100, // lets be reasonable about not making too many requests in parallel
  handler: (hook) => this.handleHook(hook)
});
```
Again, I would set `nextScheduledDate` to `new Date(MAX_INT)` if this hook isn't to be scheduled... Or something like that...
Actually, stackoverflow says `new Date(8640000000000000)` we can probably make that constant somewhere :)
see: http://stackoverflow.com/a/11526569

hooks/scheduler.js
```diff
@@ -92,11 +97,12 @@ class Scheduler {
 
   /** Handle spawning a new task for a given hook that needs to be scheduled */
   async handleHook(hook) {
-    let task = await hook.taskPayload();
-    let resp = await queue.createTask(hook.nextTaskId, task);
+    // TODO: (when we have hook logging) if this fails due to 401, we should
+    // still consider it scheduled
+    await this.taskcreator.fire(hook, {}, {taskId: hook.nextTaskId});
```
jonasfj: provide `created` see other comments on it:
`await this.taskcreator.fire(hook, {}, {taskId: hook.nextTaskId, created: hook.nextScheduledDate});`

hooks/scheduler.js
```diff
@@ -92,11 +97,12 @@ class Scheduler {
...
     await hook.modify((hook) => {
-      hook.nextTaskId        = slugid();
-      hook.nextScheduledDate = utils.nextDate(hook.schedule);
+      hook.nextTaskId        = slugid.v4();
+      hook.nextScheduledDate = nextDate(hook.schedule);
```
jonasfj: Use `taskcluster.slugid()` (gives you a nice one... and then everybody uses the same method for taskIds)...

More notably... I'm wrondering if it wouldn't be sane to do:
```js
let taskId = hook.nextTaskId;
hook.modify(hook => {
  if (taskId === hook.nextTaskId) {
    hook.nextTaskId = taskcluster.slugid();
    hook.nextScheduledDate = ....
  }
});
```
I guess it's not critical... but if two schedulers are racing (a real senario), this would have some effect if the next time schedule extremely close... or there is some clock drift.. otherwise, this wouldn't matter and is just a sanity thing...

routes/v1.js
```diff
@@ -3,11 +3,25 @@ var debug       = require('debug')('hooks:routes:v1');
...
-var utils       = require('../hooks/utils');
+var nextDate    = require('../hooks/nextdate');
 
 var api = new base.API({
   title:         "Hooks API Documentation",
-  description:   "Todo",
+  description:   [
+    "Hooks are a mechanism for creating tasks in response to events.",
+    "",
+    "Hooks are identified with a hook-group ID and a hook ID.",
```
jonasfj: nit, other places we typically write in camelCase for identifiers: `hookGroupId` and `hookId`...

hooks/data.js
```diff
@@ -8,19 +8,27 @@ var _           = require('lodash');
...
     hookId:             base.Entity.types.String,
     metadata:           base.Entity.types.JSON,
+    // task template
     task:               base.Entity.types.JSON,
+    // pulse bindings
     bindings:           base.Entity.types.JSON,
+    // timings for the task (in fromNow format, e.g., "1 day")
     deadline:           base.Entity.types.String,
     expires:            base.Entity.types.String,
+    // schedule for this task
```
jonasfj: I think this is a good place to document the form of the JSON object... at least if it's not trivial..
Or just say what sub-schema it follows... ie. some property of some schema...

routes/v1.js
```diff
@@ -126,34 +147,25 @@ api.declare({
...
+    "",
+    "The caller's credentials need not satisfy `hook.task.scopes`. Instead,",
+    "the role for the hook must satisfy these scopes as well as the necessary",
+    "scopes to add the task to the queue.",
   ].join('\n')
 }, async function(req, res) {
-  var hookGroup = req.params.hookGroup;
+  var hookGroupId = req.params.hookGroupId;
   var hookId    = req.params.hookId;
   var hookDef   = req.body;
```
jonasfj: I think we should use `deferAuth: true` and do the `if (!this.requires({hookGroupId, hookId})) { return; }` dance... 
Example: https://github.com/taskcluster/taskcluster-secrets/blob/da448a02e2639f7bb18ab4fc87bd69d0067cacd7/lib/api.js#L39-L41

routes/v1.js
```diff
@@ -163,18 +175,19 @@ api.declare({
         routingKey:  '#'
       }
```
jonasfj: Bindings was supported to be an array... but let's just give it a default value of `[]` in the JSON schema :)

jonasfj: Specifically:
```yaml
  bindings:
    type:     array
    default:  []
    items:
      type:                 object
      properties:
        exchange:
          title:            "Exchange"
          type:             string
        routingKey:         {$const: routing}
      additionalProperties: false
      required:
        - exchange
        - routingKey
```

Also see that schemas needs `description` attributes... nit, and not critical yet, but we should write a few small essays in there before we encourage people to use it :)

routes/v1.js
```diff
@@ -182,7 +195,7 @@ api.declare({
       throw err;
     }
     return res.status(409).json({
```
jonasfj: Not good enough, we have to load the existing one and check that if it matches... if it matches, then we assume previous request failed, and continue as if we were there... Idempotency...

routes/v1.js
```diff
@@ -248,22 +263,22 @@ api.declare({
...
-    "Remove a hook definition."
+    "This endpoint will remove a hook definition."
   ].join('\n')
 }, async function(req, res) {
-  var groupId = req.params.hookGroup;
+  var hookGroupId = req.params.hookGroupId;
   var hookId = req.params.hookId;
 
   // Remove the resource if it exists
   let hook = await this.Hook.load({
```
jonasfj: No need to load it...
Just do:

`this.Hook.remove({hookGroupId, hookId}, true /* ignore if not exists*/ );`

If hook doesn't exist the operation is idempotent by saying you deleted it... Otherwise, retries wouldn't be sane.

hooks/data.js
```diff
@@ -8,19 +8,27 @@ var _           = require('lodash');
...
     schedule:           base.Entity.types.JSON,
+    // access token used to trigger this task via webhook
     accessToken:        base.Entity.types.SlugId,
```
jonasfj: let's rename it `triggerToken` :)

jonasfj:
Okay, you have some comments... some of them also nits...

IMO feel free to refile this PR against a empty branch... so we have no `-` diffs... the first pass was never reviewed either...

djmitche:
Thanks!  Lots of that was just copied/modified from what Anthony wrote, so
I'll fix up the JS style a bit.  Thanks for the tips :)

hooks/taskcreator.js
djmitche: What's the allowed skew on `created`?

djmitche:
I think that addresses all of your comments.  I'll re-PR this from the empty branch.

jonasfj:
fyi, drift allowed is about 15 min:
https://github.com/taskcluster/taskcluster-queue/blob/master/routes/v1.js#L171

That seems reasonable to me... So if hooks crashes and doesn't recover in 15 min, hooks will have to be manually scheduled by button click... Seems reasonable...  Rather than having duplicates...

djmitche:
It'll be better once we mark 4xx createTask failures as failures rather
than retrying them on the next loop.

jonasfj:
> It'll be better once we mark 4xx createTask failures as failures rather, than retrying them on the next loop.

Yeah, we shouldn't retry 4xx errors... those are failures...

PR #2, opened 2015-10-21
First draft of hooks service

djmitche:
This only supports scheduled hooks -- pulse bindings and web hooks will come later (although there is some code in place already).


jonasfj:
See comments in: https://github.com/taskcluster/taskcluster-hooks/pull/3

djmitche:
OK, I merged those and fixed the syntax errors.  However, the tests don't pass, I think because they need more or different credentials.  Can you create a template for the test config that will run this successfully?

jonasfj:
will take a look at it... hopefully later today...

djmitche:
@jonasfj what's your feeling about this at this point -- good to land?

The major unresolved issue I see is the handling of scopes:
 * create task using `assume:hook-id:<hookGroupId>/<hookId>` (current implementation; represents a kind of privilege escalation)
 * require that the caller satisfy `hook.scopes` which satisfies `task.scopes` and `queue:create-task:<provisionerId>/<workerType>`, for any operation that might alter `hook.scopes` or `task.scopes`.

In the absence of a good reason to implement the former, I'd prefer to ship the latter, as it avoids the escalation path.

jonasfj:
@djmitche,

I suggest that create hook requires:
A)
 * `assume:hook-id:<hookGroupId>/<hookId>`, and
 * `hooks:create-hook`
or:
B)
 * `assume:hook-id:<hookGroupId>/<hookId>`, and
 * `hooks:create-hook:<hookGroupId>/<hookId>`

And that we use the same pattern for update, delete, etc...

When the task for a hook is created we do so with:
  `authorizedScopes: ['assume:hook-id:<hookGroupId>/<hookId>']`.

This way we have no privileged escalation. I think (B) is a bit overkill, and that (A) is a pretty solid approach for doing this. Scopes will be managed by roles, and you can't modify if you can't assume the associated role (regardless of whether or not the role exists).

----
We'll probably create a role `hook-id:releng/*` with the set of scopes that releng hooks needs, and give everybody in ldap group releng the scopes: `assume:hook-id:releng/*`, `hooks:create-hook`, etc. Possibly along with scopes required to modify roles under the `hook-id:releng/` prefix.

Or something like that, I think it's sane.

jonasfj:
@djmitche, what do you think... The changing scopes in hooks to follow pattern (A) should be trivial.

jonasfj:
@djmitche, Regarding rollout, I suggest we agree on (A), do that and roll it out...
I don't see any other things blocking us at this point. And scopes isn't really blocking us, it's just easier to do it now, and the change is trivial.

djmitche:
I'm leery of this tendency to require roles for certain actions, particularly when we do not expect the caller to necessarily have that role (as distinct from expecting the caller to have scopes that satisfy the role's scopes).

To my mind, possessing `assume:foo` means that the caller has that role: `assume:moz-tree:try` means that the call is on behalf of a push to the try tre;e `assume:hook-id:hg/h` means that the caller is hook `hg/h`.

The arrangement with the queue makes some sense, as we expect e.g., `queue.claimTask` to be called by a worker, which is what requiring `assume:worker-type:..` means.

But for hooks it makes less sense, since the creator of a hook is most likely NOT a hook -- so why would it have scopes suggesting that it is a hook?

Concretely, if `hook.task.scopes` is `['docker-worker:cache:nightly-cleanup']` and `assume:hook-id:hg/h` is `['queue:create-task:cleanup-worker/cleanup-worker']`, and the caller has `docker-worker:cache:*` and `queue:crate-task:*`, then the caller should be able to create the task.

I realize this is a little harder to implement, though.  We need a way to ask if scope-set S satisfies role R, which means fetching the role's expandedScopes and performing a `scopeMatch` operation.

jonasfj:
> But for hooks it makes less sense, since the creator of a hook is most likely NOT a hook -- so why would it have scopes suggesting that it is a hook?

Fair point. Option (B) is less like that...But yes, it still sort of violates this principal.
I actually agree here... But I'm tempted to say that "assume:..." is also just a scope one can prove that one has... Though it absolutely is intended to imply a role that can be _assumed_.

I'm not entirely sure, but I think one could also read `assume:hook-id:...` as the authority to _assume_ the role of that hook. Obviously and tasks triggered by that hook would have that authority, but I don't think it necessarily wrong to also read it as a person should _assume_ the authority of the hooks he claims to own and have the authority to modify.

> Concretely, if hook.task.scopes is ['docker-worker:cache:nightly-cleanup'] and assume:hook-id:hg/h is ['queue:create-task:cleanup-worker/cleanup-worker'], and the caller has docker-worker:cache:* and queue:crate-task:*, then the caller should be able to create the task.

So this would include scope elevation and not push all scopes to be in the role.
As this wouldn't work with authorizedScopes: [`assume:hook-id:hg/h`], because you checked the caller to see if the caller had `hook.task.scopes`. That is a bad idea.
And still scope elevation.

-------
So I see two options, both involve creating the tasks with `authorizedScopes: ['assume:hook-id:<hookGroupId>/<hookId>'].`:
1) We allow the subtle scope elevation, by not doing (A) or (B), but just requiring `hooks:create-hook:<hookGroupId>/<hookId>`
2) We prevent scope elevation with (A) or (B)

I think the scope elevation is so small and hard to exploit that option (1) is acceptable. But we might want to consider doing (A) or (B) because there actually a few places where this is tempting:
  * aws-provisioner shall give workers `assume:worker-type:aws-provisioner-v1:<workerType>`,
  * auth could require you to have the role you're modifying (should probably require you to have
    the scopes your adding, one of those scopes could be a role),
  * hooks as discussed here



So we might want to sort of agree on a pattern here as there is going to be more than one place where this concept applies. Where requiring you to have the authority to _assume_ a role, that isn't necessarily you to modify something that then assumes the role isn't bad.

djmitche:
OK, let's go with B.  To me that suggests "you must be able to edit this hook, and also have the scope the hook will run with (which happens to start with `assume`)".  Although modeling the escalation in (1) isn't hard, it's probably a better principle to just avoid scope escalation issues everywhere.

So the pattern is that editing (in an administrative sense: creating hooks, editing roles, editing workerTypes) something which will assume a role always requires that role itself be in the caller's scopes.  Where do we write down these conventions?

djmitche:
OK, that's implemented.  Just working on getting the tests to pass now..

djmitche:
I think that addresses all of the comments.  Since this isn't even deployed yet, I'm going to go ahead and merge to master.  We can address any other issues with subsequent PRs.

PR #3, opened 2015-10-21
Jonasfj review comments

jonasfj:
too lazy to write comments...

But you can consider each edit here a self documenting review comment :) he he...

I didn't run the tests... so might be some credentials issues...

Notable changes was using SAS creds from azure to talk to azure table storage... That way you just need a scope for the `jungle` account... It's intended for garbage and maximum abuse...


djmitche:
Running `npm test` isn't really that hard.. lots of syntax errors here!  I'll get 'em cleaned up.

Is this good to go, with these updates?

djmitche:
OK, merged

PR #4, opened 2015-10-23
Jonasfj review comments

jonasfj:
Okay, this is semi crazy...
But I really don't like our old config loader... it's so bad...

And I hate the duplication between files in `bin/`...
So in fixing this I took the liberty of checking out `taskcluster-lib-loader` and added a PR so we can easily inject Mock objects:
https://github.com/taskcluster/taskcluster-lib-loader/pull/3

I'm not really a fan of mocking... but I guess others are and it certainly makes sense for aws-provisioner.

I haven't got all test passing... There is also an issue with the $ref, and "default" keyword... This makes me sad, but generally I'm not a fan of $ref... It's sketchy everywhere, and complicates both validators and documentation tooling..

Anyways, we can do more to make this better... but I think this is a good start...


test/helper.js
```diff
@@ -3,58 +3,42 @@ var data        = require('../hooks/data');
...
-var bin = {
-  server:             require('../bin/server')
-};
+var load        = require('../bin/main');
+var config      = require('typed-env-config');
+var _           = require('lodash');
 
-var testProfile = 'test';
+var loadOptions = {profile: 'test', process: 'test-helper'};
+var cfg = config({profile: 'test'});
```
jonasfj: I can't use `load('cfg', loadOptions)` because it returns a promise...

I suggest that we do what I've done in other modules, make a single global `mocha.before()` hook... So we don't restart the server between suites.. We can still call `await helper.Hook.scan({}, {handler: hook => hook.remove()});`  before each test case... That way all persisted state in the api server is reset...


--------
Also I see how MockTaskCreator is nice.. but honestly for this module I wouldn't have done anything by integration tests, as it's really the only thing we care about...
Hence I would give the tasks a special route `dummy.hooks-test.<slugid>` and just listen on pulse to see them be created... 
`base.testing.PulseTestReceiver` is good for this...
See: https://github.com/taskcluster/taskcluster-base/blob/master/testing.js#L32-L49

I normally configure `base.testing.PulseTestReceiver` in the helper and expose it on the helper module.
I know this is ugly, but is it really worth the effort to write anything more than integration tests when our projects are this simple...?

I like the integration tests because they mimick production the closest, and they remain stable as long as the API is stable. Anyways, that why I don't mock things and write unit-tests, but more like black box integration tests...

I see how auth needed the unit tests especially with the dfa stuff... I also see how aws-provisioner needs to mock things because cleanup is expensive.

-----
Anyways, this doesn't need azure creds anymore... just tc creds with scopes for the `jungle` azure storage account :)
The jungle account is intended for abuse...

djmitche: If I'm reading your middle comment correctly, I'm hearing you say that the elegant DI mock, which works perfectly well and isn't dependent on external services, should be replaced with something that is "ugly" and has all the downsides of integration tests?  That doesn't make a lot of sense.  In this case, yes, writing unit tests was *considerably* easier than setting up the "ugly" mechanics you suggest, and depending on pulse (which is known to be flaky) for my tests.

You'd make a much stronger case if all of the taskcluster components had test that passed in the github-linked CI and for which contributors could easily run the tests.  Until that point, I'm going to remain firmly unconvinced.

djmitche:
I'm sorry, this is a lot of change and I don't really see what it's fixing.  Other than combining the two `bin/*.js` scripts.

test/helper.js
jonasfj: > If I'm reading your middle comment correctly, I'm hearing you say that the elegant DI mock, which works perfectly well and isn't dependent on external services, should be replaced...

Nope, lets not replace it... 

Sorry, I was just saying how I normally would do
.. And yes, it would be ugly :)

jonasfj:
> I'm sorry, this is a lot of change and I don't really see what it's fixing.

You said it was hard to get the config right...
In fixing that II went for using the new config loader that'll be part of tc-base.... I so tired of the old one...

It does fix using tc-creds only, no need for azure creds... Which makes the service more secure.. Also fixed influxdb stats, and encryption/signing for azure table...

Copied in aa $ref to make default work, see comment...


------
But yes, we could stick with the old config and the two binary files... I just figured that if I was to config easier I might as well try to do it right :)

bin/schedule-hooks.js
```diff
@@ -1,87 +0,0 @@
...
-    credentials:  cfg.get('taskcluster:credentials'),
-    authBaseUrl:  cfg.get('taskcluster:authBaseUrl'),
-    signingKey:   cfg.get('hooks:tableSigningKey'),
-    cryptoKey:    cfg.get('hooks:tableCryptoKey'),
-    process:      'schedule-hooks'
-  });
-
-  // Create a validator
-  debug("Waiting for resources to be created");
-  await Hook.ensureTable();
```
jonasfj: Actually, when using sas for azure table, issued by auth we don't need to ensure table... In fact we can't...

jonasfj:
@djmitche, yeah a lot of change we could have avoided... 

Quick question, but isn't the structure nicer?
(Particularly how the Mock injection works now)

test/helper.js
```diff
@@ -68,22 +52,20 @@ helper.setupApi = function() {
...
+    helper.Hook = await load('Hook', loadOptions);
+
+    // Remove all entities before each test
+    await helper.Hook.scan({}, {handler: hook => hook.remove()});
 
     helper.creator = new taskcreator.MockTaskCreator()
-    webServer = await bin.server(testProfile, {taskcreator: helper.creator});
+    webServer = await load('server', _.defaults({
+      Hook: helper.Hook,
+      taskcreator: helper.creator,
```
djmitche: Yes, this is pretty neat.  The application structure was already pretty nicely set up for dependency injection, but the loader makes it *much* easier.

user-config-example.yml
```diff
@@ -0,0 +1,13 @@
+defaults:
+test:
+  # Credentials to be specified for tests to work
+  # notice that you can also specify these by environment variables
+  influx:
+    connectionString: '...'
+  taskcluster:
+    # should have the scopes:
+    #  - 'auth:azure-table-access:jungle/*'
+    #  - 'queue:create-task:no-provisioner/test-worker'
```
djmitche: It also needs the scopes mentioned in `test/taskcreator_test.js

djmitche:
Yes, the structure is nicer!

So once the tests are working, I'm happy to merge this in..

test/helper.js
jonasfj: Yeah, it wasn't so bad before... I just don't like special cases like:
```js
var creator = options.taskcreator || new taskcreator.TaskCreator({
        credentials:  cfg.get('taskcluster:credentials'),
  });
```
and 
```js
  // Start scheduler
  if (!options.noStart) {
    scheduler.start();
  }
```
Last one might not be fixed yet... but I think it's easily done by splitting it in two components `scheduler` (not started) and `schedule-hooks` (started, depends on `scheduler`), well something like that...

jonasfj:
@djmitche, have look the tests should pass now... Just fill in credentials as outlined in the `user-config-example.yml` file...

jonasfj:
I think this made it into master already...

PR #6, opened 2015-11-06
Added travis badge to README and added a LICENCE

petemoore:
I see we had a travis config, but travis wasn't enabled, so I enabled it. It looks like maybe we also need to check in an encrypted config file to get tests to pass in travis.

I tried to also enable coverage reporting, but hit some issues which with my basic node knowledge I wasn't able to swiftly resolve, so instead just added a TODO here for us to enable that also.

Added a LICENCE file too, to be "proper". =)

So still TODO are:

* Get encrypted config checked in for enabling tests to pass in travis
* Enable the coverage reporting
* Optionally, cause builds to fail if coverage dips below _x_%
* Maybe add irc notifications to #taskcluster-bots IRC channel (note: @rail found a super cool solution to prevent forks from reporting in IRC [here](https://github.com/mozilla/build-tools/commit/c13a2978e0869219295013eeb52cf52227ab45ec))


djmitche:
That's why I didn't enable Travis - I couldn't get the 'travis' command-line tool installed.

djmitche:
Can you remove the Travis badge until that's complete?

petemoore:
If you send me the credentials config file gpg encrypted with my gpg public key, I can decrypt and then encrypt for travis.

petemoore:
I've added an encrypted config file and updated instructions in README.md.

Please note there are [6 skipped (pending) tests](https://travis-ci.org/taskcluster/taskcluster-hooks/builds/90487774#L1561) at the moment so we might want to work out why that is before closing this PR - perhaps more credentials are needed than are currently documented in [user-config-example.yml](https://github.com/taskcluster/taskcluster-hooks/blob/master/user-config-example.yml)?

Maybe @jonasfj might know what is needed - I think the new `user-config.yml` format is related to the new taskcluster-base loader.

Please also note, parallel test runs interfere with each other since they connect to the same endpoints and affect the same test resources. For this reason, when travis runs a pull request job at the same time as a push job, tests fail. Similarly, if we specify more than one node version, the tests can fail, due to interfering with each other. As a workaround for now I have reduced to one node version (0.12) and disabled travis jobs for pushes, keeping only jobs for pull requests. At some point, we should fix this so that the tests operate on unique resources per running test job.

djmitche:
Yeah, all of this is why I didn't include `.travis.yml` -- lots of issues with doing integration tests this way.

petemoore:
Indeed. But I think it is all fixable (e.g. if we generate random names for the test objects created). The alternative is stubbing the Queue and Auth services, which has the downside of needing maintenance and potentially being out-of-sync with production. Probably overall this current approach is good, we just need to fix the tests not to interfere with other parallel tests, e.g. adding random prefixes to things etc...

I still think we can land this once we've got those 6 tests not skipping - I think it would be bad not to have a running CI for this stuff.

djmitche:
Sounds like there's a bit more work to do, then

petemoore:
Running `npm install mocha-unfunk-reporter --save-dev; ./node_modules/.bin/mocha -R mocha-unfunk-reporter -O 'reportPending=true' test/*_test.js` I was able to see the following 6 tests are the pending ones:

```
      getTriggerToken
         returns the same token.. pending
      resetTriggerToken
         creates a new token.. pending
      triggerHook
         should launch task with the given payload.. pending
      triggerHookWithToken
         successfully triggers task with the given payload.. pending
         should fail with invalid token.. pending
         fails with invalidated token.. pending
```

petemoore:
@djmitche Looking at the source code for these tests, on each occasion it states:
```
     // XXX disabled for the first draft of this service
     this.pending = true;
```
Therefore it looks like these tests are not pending due to missing config, rather that they were never implemented. Therefore I think we should be ok to merge this now - the reason we held off before was thinking we were missing config - but now we see that is not the case.

I will also re-enable travis jobs for pushes, but set number of concurrent jobs to one, so that we get coverage back - it just means travis jobs will run serially, but this is certainly better than not running at all. At the time I disabled travis for pushes I did not realise I could achieve the same goal of avoiding parallel runs by setting the concurrency - and this way nothing gets disabled, so is a better solution.

If you are happy with the changes - feel free to hit the merge button yourself. Thanks Dustin!

petemoore:
#djmitche Note: the test failure is purely because I foolishly created a branch on my fork and don't have the right key to decrypt the secret. That is why `/pr` fails but `/push` passes. These days I branch on `/taskcluster` directly - this PR was made a while back before I wised up. =)

djmitche:
It's fine to enable this now, but I want to get this to the point where anyone -- not just the privileged few -- can run the tests.  I'm working on the fake azure-entities support to accomplish that.

petemoore:
Awesome, thanks dustin! The fake azure-entities will be a great step forward.

PR #7, opened 2015-11-06
Cron patterns have 6 entries - update schema

petemoore:
Just fixed size of array to 6 for the 6 fields used by https://www.npmjs.com/package/cron-parser.

### DIFF
--- a/schemas/schedule.yml
+++ b/schemas/schedule.yml
@@ -5,6 +5,8 @@ description: |
     If several patterns are specified, tasks will be created at any time
     specified by one or more patterns.
 type: array
+minItems: 6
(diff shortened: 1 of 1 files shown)


petemoore:
N.B. I believe these test failures are not related to my change (it is just because this is the first PR since I enabled travis this morning).

petemoore:
Erm, now I am wondering if each entry in the array is a full cron specification, so it is an array of multiple cron entries, rather than a single cron entry with 6 parts!

petemoore:
Yeah, I'm an idiot, I just read the description. :)

Closing!

PR #8, opened 2015-11-09
fix title of the hook schedule response

djmitche:


petemoore:
LOL, only just seen this after submitting PR #9.

I think they are both still valid, and they do not conflict, so I think we can still land both of them.

PR #10, opened 2015-11-24
Bug 1225240: don't retry immediately when queue returns a 4xx error

djmitche:


jonasfj:
looks good to me...

One could argue that 500 errors should be ignored as tc-client will have retried them...

And maybe set `delayFactor: 500` for the tc-client instance used for creating tasks, see:
https://github.com/taskcluster/taskcluster-client/blob/master/lib/client.js#L50-L53

If creating tasks during a request, ie. in response to a HTTP request, then the default `delayFactor: 100` is great... as you need a fast response.. But for background processes `delayFactor: 500` is awesome, as it makes everything more robust.

djmitche:
Honestly in this case we want very little retrying, as we have other tasks to create.  If things are 500'ing, hooks will retry in 60 seconds anyway.  Retrying a single task for 30s, times say 20 tasks all triggered at the top of the hour, means that we'll spin for 10 minutes in a single iteration of the polling loop.  So I think I'll do the opposite from what you suggested!

djmitche:
@jonasfj what do you think of that?

jonasfj:
Hmm... maybe just retry 1, then? Compromise, he he :)
Honestly, I think this is fine. People shouldn't expect accuracy below one hour anyways. Which gives us 60 retries in an hour.

@djmitche, note that you're running upwards a 100 `handleHook` calls concurrently. The scan thing fetches up to the `limit: 100` you gave depending on how much azure returns and runs the handler for them concurrently. So even at 30s it would perform reasonably. But yeah, just doing either zero or 1 retry seems like a good idea.

jonasfj:
`r+`

djmitche:
Ah, I didn't realize that `scan` called its callbacks concurrently.  Anyway, I stuck with 1.

PR #11, opened 2015-11-25
Bug 1225244: send emails on hook errors, if desired

djmitche:
Tested by hand, anyway.  I'll have a separate PR out for adding the emailOnError checkbox in tools.

### DIFF
--- a/bin/main.js
+++ b/bin/main.js
@@ -8,6 +8,7 @@ var taskcreator = require('../hooks/taskcreator');
 var v1          = require('../routes/v1');
 var _           = require('lodash');
 var Scheduler   = require('../hooks/scheduler');
+var AWS         = require('aws-sdk-promise');
 // These will exist in taskcluster-base, when we move to the next version.
 var config      = require('typed-env-config');
 var loader      = require('taskcluster-lib-loader');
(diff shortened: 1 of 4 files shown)


bin/main.js
```diff
@@ -58,11 +59,16 @@ var load = loader({
...
+  },
+
   router: {
-    requires: ['cfg', 'validator', 'Hook', 'taskcreator'],
-    setup: ({cfg, validator, Hook, taskcreator}) => {
+    requires: ['cfg', 'validator', 'Hook', 'taskcreator', 'ses'],
+    setup: ({cfg, validator, Hook, taskcreator, ses}) => {
       return v1.setup({
-        context: {Hook, taskcreator},
+        context: {Hook, taskcreator, ses},
```
jonasfj: Why does the API need this?

jonasfj: If you imagine that it should be triggered by the webhook too... (and in future the pulse hook), then I think it's sane to add to the `taskcreator` instead of the `scheduler`.

djmitche: No i just added it to the wrong component -- will fix.

djmitche:
how's that?

jonasfj:
Seems good to me.. `r+`

PR #12, opened 2015-12-09
Bug 1225234: Don't require hook.task to have bogus datestamps

djmitche:


hooks/data.js
```diff
@@ -32,6 +32,37 @@ var Hook = base.Entity.configure({
...
+    schedule:           base.Entity.types.JSON,
+    // access token used to trigger this task via webhook
+    triggerToken:       base.Entity.types.EncryptedText,
+    // the taskId that will be used next time this hook is scheduled;
+    // this allows scheduling to be idempotent
+    nextTaskId:         base.Entity.types.EncryptedText,
+    // next date at which this task is scheduled to run
+    nextScheduledDate:  base.Entity.types.Date,
+  },
+  migrate: function(item) {
```
jonasfj: nit... just showing that you can do:
```js
migrate(item) {
  delete item.task.created;
  ...
  return item;
},
```
or you can do:
```js
migrate: (item) => {
  delete item.task.created;
  ...
  return item;
},
```

don't fix it... just saying I love es6 :)

hooks/data.js
```diff
@@ -32,6 +32,37 @@ var Hook = base.Entity.configure({
...
+    // remove the task timestamps, as they are overwritten when the hook fires
+    // (bug 1225234)
+    delete item.task.created;
+    delete item.task.expires;
+    delete item.task.deadline;
+  }
```
jonasfj: You'll need to return `item` here.

`return item`

jonasfj:
Once the `return item` thing is done this is good to go.

PR #13, opened 2015-12-16
Bug 1233055: update nextTaskId when changing a hook

djmitche:


jonasfj:
I'm curious... Why you reset the nextTaskId... It looks okay...

djmitche:
Because otherwise you can end up in a state where the task has already been submitted, but the next run of the scheduler tries to submit it with a different created time.

jonasfj:
> Because otherwise you can end up in a state where the task has already been submitted, but the next run of the scheduler tries to submit it with a different created time.

I think you meant to say, you could end up trying to re-submit a task with different task definition. Which makes more sense. But yes, it makes a lot of sense.

djmitche:
Well, yes, the different created time means a different definition, and thus failure :)

PR #15, opened 2016-01-07
Bug 1225239: provide some information about the last time the hook fired

djmitche:


hooks/data.js
```diff
@@ -64,6 +64,37 @@ var Hook = base.Entity.configure({
...
+    // timings for the task (in fromNow format, e.g., "1 day")
+    deadline:           base.Entity.types.String,
+    expires:            base.Entity.types.String,
+    // schedule for this task (see schemas/schedule.yml)
+    schedule:           base.Entity.types.JSON,
+    // access token used to trigger this task via webhook
+    triggerToken:       base.Entity.types.EncryptedText,
+    // information about the last time this hook fired:
+    // {error: ".."} or {taskId: ".."}
+    lastFire:           base.Entity.types.JSON,
```
jonasfj: I propose `lastFired`, sorry for being a grammar nazi... :)

djmitche: Well, it's more than just the time that it was last fired..

jonasfj: ah, that's true... nvm...

schemas/hook-status.yml
```diff
@@ -0,0 +1,55 @@
...
+    description: |
+      Information about the last time this hook fired.  This property is only present
+      if the hook has fired at least once.
+    type: object
+    oneOf:
+      - title: Successful Fire
+        description: "Information about a successful firing of the hook"
+        type: object
+        properties:
+          taskId:
```
jonasfj: Other places we've used `oneOf` we've often added a property:
`kind: enum: ['task-fired']` and `kind: enum: ['fire-failure']`

to each of them... So it's that when reading the object you don't have to test for property presence, rather you look at the `kind` property. Our golang binding doesn't take advantage of this yet, but I hope someday we will (or we'll find another way of doing it).

It's somewhat nice to have this even in Javascript.

schemas/hook-status.yml
```diff
@@ -0,0 +1,55 @@
+$schema: http://json-schema.org/draft-04/schema#
+title:   "Hook status response"
+description: |
+  A snapshot of the current status of a hook.
+type: object
+properties:
+  lastFire:
```
jonasfj: Is `lastFire` supposed to be optional? Maybe the initial value can just be an error message saying it haven't been triggered yet..

djmitche:
Regarding `oneOf` and making `lastFire` non-optional -- I wasn't really sure how best to do this, so thanks for the comments.

The next iteration, then, will make `lastFire` required and add a `result` property with values `new-hook`, `fired`, and `failed`.  That should be easier to interpret later.

Are you OK with inserting the error object verbatim, or should I be trying to pre-process it somehow?

hooks/taskcreator.js
```diff
@@ -66,10 +66,17 @@ exports.TaskCreator = TaskCreator;
 class MockTaskCreator extends TaskCreator {
   constructor() {
     super({credentials: {}});
+    this.shouldFail = false;
     this.fireCalls = [];
   }
 
   async fire(hook, payload, options) {
+    if (this.shouldFail) {
+      let err = new Error("uhoh");
```
selenamarie: lol

djmitche: I make a habit of using this for errors in testing, since it's clear (or at least easy to grep for) when debugging the tests that this is an intentional error.

djmitche:
OK - updated.  Either @jhford or @selenamarie want to have another look?

jonasfj:
> Are you OK with inserting the error object verbatim, or should I be trying to pre-process it somehow?
Error object from taskcluster-client should be safe, but no promises... I think you can take err.message.

I think an exploit would be hard...

djmitche:
If we just include message, then users will only get "Authorization failed", which isn't so useful - the scopesets and list of scopes are pretty critical to debugging such things.  I think I'll leave it as-is.

PR #16, opened 2016-01-21
Run most tests entirely in-memory unless credentials are available

djmitche:
This uses the new fakeauth as well as the new inMemory azure-entities fake, so that tests talking to those two things can run without any credentials or network access (and very fast!!).

The tests for actually creating a task *do* require credentials.  There are still credentials in the travis-encrypted file to support this.  Let's see if Travis still fails on a PR.

I had to sort of hack out the requirement for an influx connection string.  Hopefully that's OK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,5 @@ services:
 - rabbitmq
 script: npm test
 before_install:
-- openssl aes-256-cbc -K $encrypted_c76f7cd53afb_key -iv $encrypted_c76f7cd53afb_iv -in user-config.yml.enc -out user-config.yml -d
+# unpack the secrets, but only if this is not a pull request; otherwise, the tests will run with no credentials
+- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && openssl aes-256-cbc -K $encrypted_c76f7cd53afb_key -iv $encrypted_c76f7cd53afb_iv -in user-config.yml.enc -out user-config.yml -d
--- a/bin/main.js
+++ b/bin/main.js
@@ -28,7 +28,7 @@ var load = loader({
         return new base.stats.Influx(cfg.influx)
       } else {
         debug("Not loading Influx -- no connection string");
-        return null;
+        return new base.stats.NullDrain();
       }
     },
   },
--- a/bin/main.js
+++ b/bin/main.js
@@ -12,7 +12,7 @@ var AWS         = require('aws-sdk-promise');
 // These will exist in taskcluster-base, when we move to the next version.
 var config      = require('typed-env-config');
 var loader      = require('taskcluster-lib-loader');
-var tc_lib_app  = require('taskcluster-lib-app');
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -8,4 +8,5 @@ services:
 - rabbitmq
 script: npm test
 before_install:
-- openssl aes-256-cbc -K $encrypted_c76f7cd53afb_key -iv $encrypted_c76f7cd53afb_iv -in user-config.yml.enc -out user-config.yml -d
+# unpack the secrets, but only if this is not a pull request; otherwise, the tests will run with no credentials
+- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && openssl aes-256-cbc -K $encrypted_c76f7cd53afb_key -iv $encrypted_c76f7cd53afb_iv -in user-config.yml.enc -out user-config.yml -d
```
jonasfj: It's easier to just set env vars on travis... it has a UI for doing this now... then we don't have ugly things in .travis... anyways just a note... don't have to do it here it...just saying...

bin/main.js
```diff
@@ -12,6 +12,7 @@ var AWS         = require('aws-sdk-promise');
 // These will exist in taskcluster-base, when we move to the next version.
 var config      = require('typed-env-config');
 var loader      = require('taskcluster-lib-loader');
+var tc_lib_app  = require('taskcluster-lib-app');
```
jonasfj: Why not base.app, I don't think we updated this recently...

bin/main.js
```diff
@@ -22,7 +23,14 @@ var load = loader({
 
   influx: {
     requires: ['cfg'],
-    setup: ({cfg}) => new base.stats.Influx(cfg.influx),
+    setup: ({cfg}) => {
+      if (cfg.influx && cfg.influx.connectionString) {
+        return new base.stats.Influx(cfg.influx)
+      } else {
+        debug("Not loading Influx -- no connection string");
+        return null;
```
jonasfj: new base.status.NullDrain will do

jonasfj:
looks good to me `r+`

bin/main.js
djmitche: I just shipped a new version of it, and I think we are getting away from bundling everything in base, right?

.travis.yml
djmitche: I'm confused -- this is an env var.

jonasfj: Sorry, I just meant as an alternative to the using the encrypted file..

bin/main.js
jonasfj: Really, I'm not.. it's a lot easier to just include one bundle... Most changes aren't breaking anyways, so we just want to downstream components to pick them up slowly.

Actually, it doesn't really matter... Except we don't have good variable naming conventions when we include modules directly... At least we all know what `base` name is...

jonasfj: this is totally nit

PR #19, opened 2016-02-19
Make tests work on travis again

imbstack:
They hadn't been running [for a bit](https://travis-ci.org/taskcluster/taskcluster-hooks/requests) (the "missing config" lines). I think the changes when we tried to "make travis run even without secrets" made this not work for some reason. The tests seem perfectly happy without that part. Not sure if they're supposed to be passing like this or not.

If this is a valid thing to do, let's ship it. If it's not, we should figure out some other way to get travis to run the tests again.


djmitche:
Without this file, only the tests that don't require credentials run.  Specifically, it skips the tests that involve creating a new task.  It would be good to test those anyway, but I don't know what's wrong with the encrypted file on travis.

Can you point to a specific failing job?  The link you used is all green jobs..

imbstack:
They look green, but haven't actually run. Anything with "missing config" appears to not run at all. I noticed this when I submitted #18 which didn't run on travis. I guess you can't tell that now that it's been merged. I've made #20 as an example to test this. There's no travis status on the PR and on [the requests tab](https://travis-ci.org/taskcluster/taskcluster-hooks/requests) of travis, it's shown up, but is marked as "missing config".

My best guess is that this line is breaking ```.travis.yml``` parsing somehow but I haven't seen how quite yet. Let's not merge this particular request, but we should try to find what's wrong with it.

djmitche:
https://github.com/travis-ci/travis-ci/issues/2025

djmitche:
looks like 7411c49e4151072ee83bffa7f122a5799f18338f introduced an unterminated string.  Try just terminating it?

imbstack:
Oh, duh. I'll close this and do that in another request.

PR #22, opened 2016-03-24
Bug 1259203: shrinkwrap hooks

djmitche:
Yup, that's pretty simple.


jonasfj:
nice, thanks for adding cache for travis too :)

jonasfj:
`r+`

PR #24, opened 2016-06-10
Fix link to docs in README

luser:
I noticed this was broken on https://docs.taskcluster.net/reference/core/hooks , and I assume the content there is generated from this README.md.


imbstack:
Thanks! Good catch.

PR #27, opened 2016-08-04
Bug 1285022 - 500 ISE when creating hook with 0 for day-of-month in the schedule

acmiyaguchi:
This pr addresses [bug 1285022](https://bugzilla.mozilla.org/show_bug.cgi?id=1285022). An invalid schedule will cause an internal server error when generating the next date, which occurs when creating/modifying a hook. This will surface the error back to the user with a 400 bad request and the cron-parser exception message.


djmitche:
Very nicely done!

PR #29, opened 2016-12-22
Bug 1324807 - tempelating system for injecting arbitrary webhook payl…

hammad13060:
> I have implemented basic functionality for a template engine.

> It doesn't support  something like "{{payload.a.b.c}} {{payload.e.f.g}}" i.e is multiple parameters in a single string (at least for now).

> Do we need support for functions ?

### DIFF
--- a/tempelate_engine/tempelate_engine.js
+++ b/tempelate_engine/tempelate_engine.js
@@ -1,0 +1,89 @@
+/*
+*========usage=========
+*body = {
+*  name: "Jim",
+*  id: "123",
+*  token: "abc",
+*  tasks: ["enter", "exit"]
+*}
+*
+*tempelate = {
+*  username: "{{payload.name}}",
+*  clientId: "{{payload.id}}",
+*  accessToken: "{{payload.token}}",
+*  task: "payload.tasks.0"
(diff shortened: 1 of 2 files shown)


tempelate_engine/tempelate_engine.js
```diff
@@ -0,0 +1,89 @@
...
+*
+*context = {payload: body}
+*
+*engine = new TempelateEngine(tempelate, context);
+*engine.render();
+*engine.getTempelate() // returns rendered tempelete
+*
+*/
+
+class TempelateEngine {
```
djmitche: I hate to nitpick on spelling, but the "e" in the middle shouldn't be there: "Template" :)

hammad13060:
Yes we can use @jonasfj project and wrap it with a template engine which is instantiated once.

tempelate_engine/tempelate_engine.js
hammad13060: Ya sure 👍

jonasfj:
I've updated the README for https://github.com/jonasfj/json-parameterization

please have a look at the tests too...

PR #30, opened 2016-12-22
Bug 1321852 - Hooks service ISEs on triggering an invalid task

hammad13060:
@djmitche is the error message fine? I was trying to write test for invalid task definition but it seams I cannot get past the creatHook method.

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -408,7 +408,16 @@ api.declare({
     hook.lastFire = lastFire;
   });
 
-  return res.reply(resp);
+
+  if (resp) {
+    return res.reply(resp);
+  } else {
+    return res.reply({
+      result: "error",
+      error: err.toString(),
+      time: new Date()
+    });
+  }
 });
 
 /** Get secret token for a trigger **/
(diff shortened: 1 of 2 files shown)


routes/v1.js
```diff
@@ -408,7 +408,16 @@ api.declare({
...
-  return res.reply(resp);
+
+  if (resp) {
+    return res.reply(resp);
+  } else {
+    return res.reply({
+      result: "error",
+      error: err.toString(),
+      time: new Date()
+    });
```
djmitche: If you return a 200 OK response (which is what `res.reply` does), then the result needs to follow the schema named in the `output` option to `api.declare`, above.  In this case, it's `task-status.json` and it's illustrated at https://docs.taskcluster.net/reference/core/hooks/api-docs#triggerHook.

However, we don't want to return a 200 OK in this case!  The way to return a different error code is to use `res.status`, with a message.  You can see some examples elsewhere in `routes/v1.js`.

I would suggest returning a 400 status (bad request) including `err.toString` with the prefix "could not create task: ".

PR #31, opened 2017-01-08
Templating engine discussion

hammad13060:


hammad13060:
@djmitche and @jonasfj 
I have filled this PR for discussion regarding json parameterisation. We can decide on what features are required in json-e npm module.

@jonasfj has mentioned following things [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1324807#c9)

I don't think we need both {{...}} and ${...}.
  I would stick to ${...} and say that if the expression covers the full string, then the value of the
  expression replaces the string, otherwise it's string interpolation as expected. Granted making it
  always string interpolation might be more sane and predictable.

  As for {$eval: ''}, $if and $switch, I see no need to require the ${...} wrapper as those strings always
  have to be expressions, it would never make sense to do string interpolation for those.

Also:
Rather than babel-compile consider using neutrino with taskcluster preset for node. Ask Eli on irc, if
you have any problems with it. I don't think we're using it for many libs yet, but it gets you everything
pretty easily.

P.S ignore the files changed (it is because of rebasing)

djmitche:
Regarding `{{...}}` and `${...}` -- a critical advantage of this tool over mustache or something like that is that this tool can substitute whole chunks of JSON-structured data, rather than just change strings in an existing structure.  So "making it always string interpolation" is not a great option, as it means we could not use that feature anymore.

I suggest keeping the two expansion syntaxes separate, as otherwise it's easy to change something like `"foo": "[${my.expression}]"` to `"foo": "${my.expression}"` and unwittingly change what it means (suddenly my.expression isn't interpolated anymore).

I had made some comments regarding the API, too (that it need not be a class, but could be a function).  I'd like to see those addressed, either with comments as to why this is a better way to do it, or by changing the implementation :)

hammad13060:
1. well I opted for different syntaxes for string interpolation (```{{...}}```) and javascript object substitution (```${...}```) because someone reading the template code will be able to perceive whether user is going for string interpolation or replacing string with javascript object.

2. @jonasfj yes I agree that ```$if``` and ```$switch``` will always evaluate an expression so using ```${...}``` is not necessary, but to the user of this tool who will constantly use ```${...}``` for evaluation at other places (replacing string with js objects), so having same syntax everywhere may be better.

3. As far as ```{$eval: ''}``` is concerned, I think this feature is redundant because it provides the same functionality as ```${...}```

4. I agree @djmitche that current api is a bit too much of work for the user, maybe making it a function is a better option.

jonasfj:
Hmm, I propose `${...}` for string interpolation and `{$eval: '...'}` for
object replacement.
And no string syntax for object replacement.

Maybe it's a different keyword than $eval, but same form... An object is
replaced by an object.

Den 9. jan. 2017 8.13 AM skrev "Hammad Akhtar" <notifications@github.com>:

jonasfj:
That way `${...}` is similar to how JS strinh interpolation works... And
object injection is clearly signaled with the $eval object.

Note: this way anything that changes the shape of the JSON object structure
has the form:  `{$<keyword>: 'expression', <options>...}`
So your never in doubt if something changes the structure, or if it just
interpolates a string.

Den 9. jan. 2017 4.56 PM skrev "Jonas Finnemann Jensen" <jopsen@gmail.com>:

jonasfj:
@hammad13060,
In talking to @imbstack  today I realize that we might have a few different use-cases for `json-e`.
Ie. other places where we want to do this kind of JSON transformation.

bstack wondered if perhaps we could come up with an expression language that could be easily implemented in python and javascript, instead of using safeEval.
I suspect it would be somewhat easy to make a safe expression based language interpreter using pegjs...
At least for javascript. If we think about it what do we really need:
 * Array access `my_array[integer]`
 * Dictionary access `mydict[key]` and `mydict.key`
 * Array slicing `my_array[start:end]` (like in python)
 * Function calls `my_func(param1, param2)`
 * String concatenation `a + b`
 * Basic math operators
 * Literal integers and strings (with escape sequences)

Most other things like length-of-array, joining-an-array, etc. can be implemented using functions.

Anyways, maybe that's crazy... Or maybe we can just start with the simplest possible expression based language, and then add features as we go.

---
@hammad13060, by the way, can't you just file a PR from a branch on the json-e repo to an empty branch on the json-e repo. That way we can comment on the json-e  stuff..

djmitche:
What were those other use-cases?  It would be great to have some concrete inputs, so we can be sure that the result is expressive enough for those inputs.

In general, I have been thinking about hook bodies (set up to handle either pulse messages or webhook payloads), `.taskcluster.yml`, and an "enhanced" task creator that can be started with a link to a template and default input.  I agree that we should use the same engine for all three!

I think we should draw a line ("make the easy stuff easy, and the hard stuff possible") between things that can happen in a template and things that require a decision task.  So I like that the previous comment does not include any loops or the possibility of recursion (note: conditionals might be nice) as that makes it very "safe" against malicious use.

Note: XST may be a good model for this.  There's a little prior art:
 * https://www.npmjs.com/package/jsonpath-object-transform
 * https://github.com/bazaarvoice/jolt

Anyway, let's ground this conversation in some concrete use cases.

imbstack:
I only brought this point up because if this is used in Action Tasks of the Future™, Treeherder will need to interface with it somehow. At the moment, we're going to try to do that all in frontend JS, but for now it's done in Python by pulse_actions. That's the specific point where I thought there might be some issues with tying it so closely to Javascript. From a general point of view as well, I figured if we're specifying some new part of our platform, it may as well work in whatever languages people want, so long as they're willing to put in the effort.

hammad13060:
@jonasfj 
regarding $eval
I feel that we already have to use curly braces allot while coding in javascript (this sometimes causes lots of confusion if we have callbacks within callbacks, thus making debugging difficult). Thus having object replacement construct being enclosed by another pair of opening/closing curly braces may not be a good idea (template may look untidy).

@djmitche  @imbstack and @jonasfj 
 Its great that we have requirement for something like json-e at other places, maybe we can test it by using it in actual tasks.

As far as the idea of common expression language is concerned. It seems like a good idea (common expression language for evaluation in different languages), but I suspect it seems like a big project in itself because I feel we will like to develop something as powerful as safeEval at least.


As far as the json-e is concerned I propose following things are enough (at least for now)
1. ```{{...}}``` for string interpolation
2. ```${...}``` for object replacement
3. ```$if``` construct
4. ```$switch``` construct
5. recursive evaluation of nested objects

jonasfj:
As far as the idea of common expression language is concerned. It seems
> like a good idea (common expression language for evaluation in different
> languages), but I suspect it seems like a big project in itself because I
> feel we will like to develop something as powerful as safeEval at least.

Yeah, it's a non-trivial thing... I suspect we could do a simple
interpreter using PEGs, that would be easy to port between languages with
PEG libraries.
Notably just having, obj[key], array[index], obj.key to select from an
object structure + function invocation and basic operators would get us
very far, and be fairly simple using PEGs.
Then maybe later we can add literals, the only hard part here is strings,
other literals is easy.
But things like ternary-if-expressions, inline-functions, etc. is very
hard, but I also doubt such things are necessary.

Mostly we just want expressions  that given a JSON structure can select
some value.
Ie. expressions like: `myobj.myarray[5]`
Doing function calls and inserting literals like: `myfunc(myobj.myarray[1],
"some string")`
And combing/comparing things with operators...

We probably don't need or want: inline-function definition, loops, custom
variables, assignments, modifications of values in the context, and things
like that.
If we tried we could probably do the basics in PEG.js in a few hours or
so...

----
@hammad,
I really don't like having two syntaxes like `{{}}` and `${}`, either:
 A) one can do both, interpolation and replacement, or,
 B) we only have string interpolation and require $eval for replacement.

Note: that in practice we're not going to be writing this in JSON, when
humans write it they are mostly likely going to be writing it in YAML,
which then gets translated to JSON. We've found that JSON is not fun to
write :) We humans always forget the commas.

Example:
```yaml
MyObject:
  MyKey: {$eval: 'some expression'}
  MyKey2:
    $eval: 'some other expression'
  MyKey3:
    $eval: |
      some multi-line
      expression, just because we can have long expressions,
      though I doubt we'll need this much...
  MyOtherKey:
    $if: 'some-expression'
    then: 'some-string-in-case-of-true'
    else: [1,2,3] # a list of numbers in case of false
```
Notice that YAML also have comments, another reason to use that over YAML,
at-least for humans :)
(in automation we always convert to JSON, but often config files are in
YAML)


--
Regards Jonas Finnemann Jensen.

2017-01-12 4:48 GMT-08:00 Hammad Akhtar <notifications@github.com>:

jonasfj:
@hammad13060,
Okay, I got myself distracted playing with peg.js :) hehe...
The following is a simple expression based language interpreter:
https://gist.github.com/jonasfj/c0f807d8410ad54e54f77e04724bbf74

Try it out with peg.js: https://pegjs.org/online
I always use the online test site for development.

Clearly we need to some test cases, and we need to make sure I got the the operator precedence stuff right.. I'm sure there is a bug or two in there... 
It's also possible that a few hacks in necessary to ensure that we test `hasOwnProperty`, and prevent access to built-in javascript properties that could otherwise be used to access window and by implementation `window.eval`.. It's probably not trivial to do, but if the context is `{myobj: {key: 7}}`, then maybe there is fun things people can do with `myobj.prototype.` or `__proto__`, stuff that `hasOwnProperty` would prevent.. And maybe a few other things who knows.

As for implementing this in python, I'm not exactly sure how to go about it, but there is quite a few PEG implementations for python, one of them must be easy to use :)

I'm sure this isn't perfect, but It'll get us very far... Maybe in the future we would want to ensure that more string escape sequences works, or more character classes works as identifiers... Or who knows maybe someone wants bitwise operators too. IMO, we can add such things later, IMO this is something we realistically implement in a few different languages.

djmitche:
Let me try to introduce a use case or two here, to be concrete.

I want to schedule a task every time a task is resolved as an exception,
https://docs.taskcluster.net/reference/platform/queue/references/events#
taskException
I only want to do this when reasonResolved for the most recent run is
intermittent-task.
That is, `message_payload.status.runs[-1].reasonResolved ===
"intermittent-task"` (assuming a syntax where negative array indices are
relative to the end of the array).  I want to include the taskId as an
environment variable in the resulting task definition.

I am setting up a webhook with bitbucket, for [push events](
https://confluence.atlassian.com/bitbucket/event-payloads-740262817.html#
EventPayloads-Push) and I would like to add one item to `task.command` for
each entry in the push.changes array, containing the branch's revision hash.

Would this language be sufficient for both of those?  Are there other cases
you can think of that would be more complex?

djmitche:
Also, if we can avoid it, I'd rather *not* allow $eval.  The danger of
executing arbitrary code within our server instances is great --
process.env contains secrets, a while-true loop could easily cause
downtime, and so on.  If we can't avoid it, I'd like to use something
like the `vm` library that Jonas used in json-parameterize, as well as
common good practices like using `hasOwnProperty`.

jonasfj:
things like extending to support negative array indices and array slicing would be pretty easy...
It probably wouldn't be too hard to add ternary if-expressions, and 

@djmitche, the idea with `{ $eval: 'expr'}` is to use the expression-based language interpreter to interpret `'expr'`, and then replace the `{$eval: 'expr'}` object with the resulting value.
Hence, say we have context: `{mykey: [1,2,3], mystr: 'hello'}`, the way to substitute `[1,2,3]` into a JSON object, is to write:
```js
{
  "array_from_context": {"$eval": "mykey"},
  "string_interpolation": "${mystr} world",
}
```
instead of having:
```js
{
  "array_from_context": "{{mykey}}",
  "string_interpolation": "${mystr} world",
}
```
as having both `'${...}'` and `{{...}}` with one being for object replacement, and the other for string interpolation is super confusing.

Both `"${...}"` and `{$eval: "..."}` will evaluate `...` using the expression-based language interpreter.
Instead of `{$eval: '...'}` we could also call it `{$subst: '...'}`, as in substitute the object for the value of an expression.


----
For many advanced things we can inject functions into the context. Slicing, negative indexes don't have to be supported by the interpreter, we can just inject functions through the context. Obviously, that's not ideal, as some things are easier to express with syntax instead of functions.

As for things like `.map` and `.filter` we can possibly introduce inline functions as long as they only contain a single expression that should be fairly simple. We could also do this in the JSON syntax using something like:
```js
{
  myarrary: {
    $for: '<variable_name>',
    in: '<expression that evaluates to an array>',
    each: { // JSON object where context contains variable_name as an element from the array
      revision: '${<variable_name>}',
    }
  }
}
```

Obviously, it would be more powerful to have this as part of the expression language.

djmitche:
I see -- I did misunderstand $eval, so thanks for clarifying.  I'm totally
onboard with what you're suggesting now!

Possible next steps:
 * build a proof of concept in JS
 * use it to implement some use cases (such as the two I mentioned above)
   * tweak the language as necessary to support those use-cases (we might
find another simple, useful feature to add, like negative array indexing)
 * write a more formal spec for the language and develop a set of test
cases that can be used to verify all implementations
 * write it again in Python so we can use it in action tasks
 * package it up and start using it everywhere!

jonasfj:
I would suggest we start with writing the test cases... A YAML file on the form:
```yaml
---
title: 'String interpolation in object value'
template: {val: 'Hello ${key}', ...}
context: {key: 77, ...}
result: {...}
---
...
```
Is a great way to:
 * test everything
 * ensure compatibility across languages
 * document all features (obviously we need both simple and complex examples)
Furthermore, agreeing on such a document would also be a good way to agree on the syntax. It's easy to discuss things this way.

In fact, if we have such a document we probably don't need to write a spec. We easily have test cases for all the syntax we want to support and all the built-in functions we want the expression language to have...

djmitche:
Sure!  Some of those test cases could represent the use-cases we expect :)

hammad13060:
@djmitche and @jonasfj It seems like a great idea :)

@jonasfj while going through your code I stumbled upon this problem, suppose we have following expression 
```1 + 9 / 3 * 2``` ideally answer should be 7 but it returns 2.5. I feel this is probably because rule for multiplication is specified before division hence it has higher precedence. Ideally both * and / should have equal precedence and in case of precedence conflict, evaluation evaluation should take place from left to right.

I am experimenting with your language @jonasfj and will keep you guys updated.

jonasfj:
@hammad13060,
Yeah, I'm sure there is a lot of subtle cases like this :)
Probably we need test cases... And lots of them, hehe..
If you look at the default example in the pegjs it uses repetition with a `expr ('*' expr)*` pattern, that might be better..

Note: We should probably make two YAML test files... One called spec.yml and tests.yml, where spec is the bare minimum cases to exsmplify all features and tests is All the arbitrary test cases we can come up with.
Note: since the expression language is embedded in json-e we can test its implementation too with these test files.

jonasfj:
@hammad13060,
Either we use safeeval + some similar python thing, but I doubt it's easy to find a safe expression based language that implemented in both python and js.

Hence, why I suggest we define our own language.. and implement in python and node...
I'm pretty sure it's easy to find a PEG library for python, just googling finds me 3, once we pick one porting to python is likely easy.

PR #32, opened 2017-01-12
Fix error handling when task cannot be created

djmitche:
Fixes:
```
ReferenceError: err is not defined
  File "/app/routes/v1.js", line 416, in Object.callee$0$0$
    error: "could not create task: " + err.toString()
```

@hammad13060 - r?

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -402,20 +402,16 @@ api.declare({
       error: err,
       time: new Date(),
     };
+    return res.status(400).json({
+      error: "could not create task: " + err.toString()
+    });
   }
 
   await hook.modify((hook) => {
     hook.lastFire = lastFire;
   });
 
-
-  if (resp) {
-    return res.reply(resp);
-  } else {
-    return res.status(400).json({
-      error: "could not create task: " + err.toString()
-    });
-  }
+  return res.reply(resp);
 });
 
 /** Get secret token for a trigger **/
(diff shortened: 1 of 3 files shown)


hammad13060:
Seems right to me, but why didn't we catch this error [here](https://github.com/hammad13060/taskcluster-hooks/blob/6d74c77169389c693d17077b66efdb042ef0c31b/test/api_test.js#L261) earlier.

Does this have to do something with the node version ??

djmitche:
I bet it's this line:
```
 (err) => { debug("Got expected authentication error: %s", err); });
```
in fact that error is *not* expected -- we should get the 400 that the lines above assert.

djmitche:
(The error that line is logging about is the ReferenceError!)

djmitche:
Actually, it's worse than that -- the triggerHook call itself returns InvalidScopes (rather than the createTask call it makes returning that error).

djmitche:
OK, test case updated too.  That got a little more complicated than I expected!

hammad13060:
this test checks for the proper reply if task trigger is not successful. Probably we will need another test which  checks for invalid scopes.

djmitche:
Oh, good point -- the name is wrong.  If hook.task had invalid scopes, we'd get this error -- or if it caused an error for some other reason.

Calling triggerTask with incorrect scopes isn't that interesting, since taskcluster-lib-api takes care of checking that.  So it's not something we need to test in hooks.

djmitche:
test name updated.

hammad13060:
Hmm, now this test looks good :)

PR #33, opened 2017-01-12
Fix missing err in error

imbstack:
@hammad13060: An error popped up in Sentry about `err` not being defined here. I've tried to make it defined without changing the initial logic a whole lot. If there is a better way to make this nice, let me know and we can do that instead. For now, if this looks good to you, we can merge it!

Error from Sentry:
```
ReferenceError: err is not defined
  File "/app/routes/v1.js", line 416, in Object.callee$0$0$
    error: "could not create task: " + err.toString()
  File "/app/node_modules/babel/node_modules/babel-core/node_modules/regenerator/runtime.js", line 61, in tryCatch
    return { type: "normal", arg: fn.call(obj, arg) };
  File "[as _invoke] (/app/node_modules/babel/node_modules/babel-core/node_modules/regenerator/runtime.js", line 323, in GeneratorFunctionPrototype.invoke
  File "function) [as next] (/app/node_modules/babel/node_modules/babel-core/node_modules/regenerator/runtime.js", line 94, in GeneratorFunctionPrototype.prototype.(anonymous
  File "/app/node_modules/babel/node_modules/babel-core/node_modules/regenerator/runtime.js", line 132, in GeneratorFunctionPrototype.invoke
    var result = generator[method](arg);
  File "internal/process/next_tick.js", line 129, in process._tickDomainCallback
```

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -382,6 +382,7 @@ api.declare({
   var resp;
   var payload = req.body;
   var hook = await this.Hook.load({hookGroupId, hookId}, true);
+  var error = null;
 
   if (!hook) {
     return res.status(404).json({
@@ -397,6 +398,7 @@ api.declare({
       time: new Date(),
     };
   } catch(err) {
+    error = err;
     lastFire = {
       result: 'error',
       error: err,
@@ -413,7 +415,7 @@ api.declare({
     return res.reply(resp);
   } else {
     return res.status(400).json({
-      error: "could not create task: " + err.toString()
+      error: "could not create task: " + (error || "unknown").toString()
     });
   }
 });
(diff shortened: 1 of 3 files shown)


imbstack:
https://sentry.io/taskcluster/taskcluster-hooks/issues/203962945/

djmitche:
See also #32, where I just used an early return.  We can merge whichever one Hammad thinks is best :)

hammad13060:
@djmitche and @imbstack 
I like this one because all the response logic for both the scenarios is at one place 
@imbstack nicely done 👍

djmitche:
Brian, do you want to cherry-pick the test fix from #32 and merge?

imbstack:
Test from #32 merged in here now. Will deploy as soon as tests pass.

PR #34, opened 2017-01-12
Remove tc-base to fix heroku install

imbstack:
Warning: Annoying review ahead

This is one part of the modernization process. It's pretty much just ripping out tc-base and updating any libraries to more modern versions. Most of the changes happen because of the update to tc-lib-validate. I did this today because it's the only way I could see to get it to compile on heroku.


bin/main.js
```diff
@@ -26,10 +26,10 @@ var load = loader({
     requires: ['cfg'],
     setup: ({cfg}) => {
       if (cfg.influx && cfg.influx.connectionString) {
-        return new base.stats.Influx(cfg.influx)
+        return new stats.Influx(cfg.influx)
```
djmitche: haha, influx

PR #35, opened 2017-01-20
Set taskGroupId = taskId

djmitche:
This follows the conventions for decision tasks.  It doesn't hurt
for non-decision tasks.


hooks/taskcreator.js
```diff
@@ -25,6 +25,9 @@ class TaskCreator {
     task.created = created.toJSON();
     task.deadline = taskcluster.fromNowJSON(hook.deadline, created);
     task.expires = taskcluster.fromNowJSON(hook.expires, created);
+    // set the taskGroupId to the taskId, thereby creating a new task group
+    // and following the convention for decision tasks.
+    task.taskGroupId = options.taskId;
```
jonasfj: For the record, this is already the default, if no taskGroupId is specific.

jonasfj: taskGroupId is an optional property

djmitche: Ok, thanks -- I didn't realize this was the default (I figured `slugid()` might be the default).  But still, explicit is better than implicit.

PR #36, opened 2017-01-21
Modernize Hooks

djmitche:
This adheres to modern standards for TC microservices.

I'm starting to [write those standards down, too](https://public.etherpad-mozilla.org/p/dustin-modernizing-microservice) and will create a docs page with the results.

r? jhford since you built a lot of these modernizations :)


jonasfj:
@djmitche please consider documenting them in the taskcluster-skeleton repository... I already started on the folder layout....

Procfile
```diff
@@ -1,2 +1,2 @@
-web:        babel-node bin/main.js server
-scheduler:  babel-node bin/main.js scheduler
\ No newline at end of file
+web:        babel-node src/main.js server
```
jhford: so this might work, but it will end up actually not using the compiled (and tested) javascript code.  Instead, what you should be doing is `node lib/main.js ...` for both entries.

config.yml
```diff
@@ -48,6 +43,9 @@ defaults:
       secretAccessKey:  !env AWS_SECRET_ACCESS_KEY
       region:           us-west-2
       apiVersion:       2014-01-01
+  monitor:
+    mock:             true
```
jhford: in other `main.js` files, we usually compare the environment to production and use that truthiness value instead of coding a `true` or `false` into the config files.  not too important.

src/main.js
```diff
@@ -22,34 +21,22 @@ var load = loader({
...
-        return new raven.Client(cfg.raven.sentryDSN);
-      }
-      return null;
-    }
+  monitor: {
+    requires: ['process', 'profile', 'cfg'],
+    setup: ({process, profile, cfg}) => monitor({
+      project: 'taskcluster-hooks',
+      credentials: cfg.taskcluster.credentials,
+      mock: cfg.monitor.mock,  // true in production, false in testing
```
jhford: this is where i'd do the `mock: profile !== 'production'` check.

test/runtests.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 # USAGE: Run this file using `npm test` from repository root
 
-mocha test/*_test.js
+mocha .test/*_test.js
```
jhford: We can safely delete this file i believe?  Its only purpose was to wrap the package.json script entry iirc, so now that you're calling into mocha directly, It's not important to exist.

package.json
```diff
@@ -13,18 +16,22 @@
   "dependencies": {
     "assume": "^1.2.5",
     "aws-sdk": "^2.2.0",
-    "azure-entities": "^0.9.0",
+    "azure-entities": "^0.9.1",
     "babel": "^5.8.21",
```
jhford: This should probably be removed because it's no longer needed as babel-compile pulls in the correct version of babel-core.

Procfile
jhford: Oh, and also, this behaviour is depending on babel-node being installed, but that binary was removed (iirc) in an update to the 'babel' package a while ago, so this would likely fail on deploy as the required binary would be absent.

djmitche: Oh, good catch, thanks!  It definitely won't work as-is!

config.yml
djmitche: It seems to go both ways (I copied this from another service).  I'll add this to my list for best practices.

test/runtests.sh
djmitche: Another good catch :)

djmitche:
@jhford I think that addresses the fixes you mentioned.  I'm away until Monday, so there's plenty of time if you want to have another look -- otherwise I can just merge on Monday given the r+ above.

jhford:
Looks good to me!

PR #38, opened 2017-01-23
fix the test of the MockTaskCreator

djmitche:
This wasn't using `await`, so while it was failing, that error was never
seen except in a log to stderr now that we run node 6..

The issue was that the hook definition lacked some fields that the fire method was returning.


imbstack:
I guess this needs the other PR to land before the tests could pass? I assume this works locally, so r+!

djmitche:
Yeah, Travis building PRs is messed up :(

PR #37, opened 2017-01-23
Use tc-lib-docs

djmitche:
This is based on #36.  I'm interested to see what Travis does with it, since there's no change to `npm-shrinkwrap.json`.  Have I shared with you my feelings about npm?

PR #39, opened 2017-01-24
bump tc-client, and fix test case

jonasfj:


jonasfj:
I don't think PRs will ever pass here... The error seems to be related to decrypting a file only avaialble for non-PRs.

djmitche:
They can succeed, just skipping the tests requiring credentials -- #37 makes that happen (but has other issues, if possibly you can be of any help there?)

PR #41, opened 2017-02-13
Bug 1332742 - Use tc-notify to send task-failure alerts; r=dustin

hammad13060:
@djmitche 
I am not able to figure out, which email should I use to replace ```*``` [here](https://github.com/taskcluster/taskcluster-hooks/pull/41/files#diff-1c90ff38a08209f9ebd4d05d1e43358eR67).


src/main.js
```diff
@@ -59,9 +60,12 @@ var load = loader({
     setup: ({cfg}) => new taskcreator.TaskCreator(cfg.taskcluster),
   },
 
-  ses: {
+  notify: {
     requires: ['cfg'],
-    setup: ({cfg}) => new AWS.SES(cfg.aws.ses),
+    setup: ({cfg}) => new taskcluster.Notify({
+      credentials: cfg.taskcluster.credentials,
+      authorizedScopes: ['queue:route:notify.email.*.on-failed'],
```
djmitche: The scope should just be `...email.*`, meaning "can email anyone".

However, note from [the docs](https://docs.taskcluster.net/reference/core/taskcluster-notify/references/api#send-an-email%0Aexperimental) that the correct scope is `notify:email:*`.

BTW, I think using authorizedScopes here is a great idea, as it prevents us from doing unexpected things with this client.  "Least privilege" is a great design principle.

src/scheduler.js
```diff
@@ -164,38 +164,27 @@ class Scheduler extends events.EventEmitter {
...
-P.S. If you believe you have received this email in error, please hit reply to let us know.`),
-            Charset: 'UTF-8',
-          },
-        },
-      },
-    }).promise();
+    this.notify.email({
+      address: email,
+      subject: utf8(`[Taskcluster Hooks] Scheduled Hook failure: ${hook.hookGroupId}/${hook.hookId}`),
+      content: utf8(`The hooks service was unable to create a task for hook ${hook.hookGroupId}/${hook.hookId},
```
djmitche: I'm pretty sure you can drop the `utf8(..)` call here and just use the string.

src/scheduler.js
```diff
@@ -164,38 +164,27 @@ class Scheduler extends events.EventEmitter {
...
-Thanks,
-TaskCluster Automation
-
-P.S. If you believe you have received this email in error, please hit reply to let us know.`),
-            Charset: 'UTF-8',
-          },
-        },
-      },
-    }).promise();
+    this.notify.email({
```
djmitche: needs `await`

djmitche:
Here's an example of injecting a fake (in this case, a github API client):

https://github.com/taskcluster/taskcluster-github/blob/master/test/helper.js#L67
```js
  let overwrites = {profile: 'test', process: 'test'};
  helper.load = async (component) => {
    if (component in overwrites) {
      return overwrites[component];
    }
    let loaded = overwrites[component] = await load(component, overwrites);
    return loaded;
  };

  // inject some fakes
overwrites.github = fakeGithubAuth();
```

so in your case you'd inject a fake with `overwrites.notify = {email: async (..) => { ..}, }`, recording the call that was made.  You might want to use [sinon](http://sinonjs.org/) but that might be more than necessary for just one fake method.  @owlishDeveloper (owlish on irc) has worked with this a little bit and may be able to answer questions about it.

src/scheduler.js
hammad13060: I don't think we need await here since we are using synchronous notify

from the [docs](https://docs.taskcluster.net/reference/libraries/taskcluster-client)
```javascript
import asyncio # Only for async 
// Create Notify client instance
import taskcluster
import taskcluster.async

notify = taskcluster.Notify(options)
# Below only for async instances, assume already in coroutine
loop = asyncio.get_event_loop()
session = taskcluster.async.createSession(loop=loop)
asyncNotify = taskcluster.async.Notify(options, session=session)

# Sync calls
notify.email(payload) # -> None`
# Async call
await asyncNotify.email(payload) # -> None
```

hammad13060:
@djmitche is this the correct approach ?

src/main.js
```diff
@@ -64,7 +64,7 @@ var load = loader({
     requires: ['cfg'],
     setup: ({cfg}) => new taskcluster.Notify({
       credentials: cfg.taskcluster.credentials,
-      authorizedScopes: ['queue:route:notify.email.*.on-failed'],
+      authorizedScopes: ['queue:route:notify.email.*'],
```
djmitche: Still the wrong scope :)

test/scheduler_test.js
```diff
@@ -19,6 +19,7 @@ suite('Scheduler', function() {
       Hook: helper.Hook,
       taskcreator: creator,
       pollingDelay: 1,
+      notify: helper.notify,
```
djmitche: Ah, this is a *much* easier kind of dependency injection than I was suggesting.  You can ignore all the stuff I said about loader and overwrites :)

Since the fake notify is only used here, you could just require fakeNotify in this file, similar to how creator is set up a few lines earlier.

test/scheduler_test.js
```diff
@@ -178,5 +179,17 @@ suite('Scheduler', function() {
...
+
+    test('on error, notify is used with correct options', async () => {
+      let oldTaskId = hook.nextTaskId;
+      let oldScheduledDate = hook.nextScheduledDate;
+
+      creator.shouldFail = true;
+
+      assume(scheduler.notify.credentials).exists();
+      assume(scheduler.notify.authorizedScopes).exists();
+      assume(scheduler.notify.email).exists();
```
djmitche: These three `assume` options are just assuming that the the fake is set up correctly, rather than checking anything about how the scheduler called it.  I like the use of `assume` in the fake itself -- that's a fine approach.

Another approach would be to record the arguments to the `notify.email`call (say in `notify.lastEmail`), then check them in the test case.  Something like:

```
  assume(scheduler.notify.lastEmail.address).to.equal('someone@somewhere.com');
```

hammad13060:
@djmitche have a look :)

src/scheduler.js
```diff
@@ -162,8 +162,14 @@ class Scheduler extends events.EventEmitter {
       errJson = `(error formatting JSON: ${e})`;
     }
 
-    this.notify.email({
-      address: email,
+    let email = this.createEmail(hook, err, errJson);
+    this.notify.email(email);
+    this.notify.lastEmail = _.merge({}, email, {err, errJson});
```
djmitche: I like that you have factored out creating the email!

However, I think that the fake `email` method should set `lastEmail`, rather than setting it here.  This is "production" (meaning, not test) code, so where possible it's best to not write testing-specific things here.

test/fake-notify.js
```diff
@@ -0,0 +1,11 @@
+let assume = require('assume');
+
+module.exports = {
+  credentials: 'test-credentials',
+  authorizedScopes: ['test:scope'],
+  email: (payload) => {
+    assume(payload.address).exists();
+    assume(payload.subject).exists();
+    assume(payload.content).exists();
```
djmitche: so, write something like `module.exports.lastEmail = payload` here.

src/scheduler.js
hammad13060: so how will you go about accessing the ```notify``` in dummy email function ?

djmitche: Something like this would do:

```
notify = {
  ...
  email: (payload) => {
    ...
    notify.lastEmail = payload
  }
};

module.exports = notify;
```

djmitche: That's just putting `notify` in a local variable so that it can be accessed from within the `email` function.

hammad13060: lol, such a silly question, I think I should take rest :p

hammad13060: ```this``` keyword should also work instead of ```notify```.

hammad13060: probably for ```err``` and ```errJson``` we don't have this option, so what should I do set them in production code or leave the testing for ```content``` of email ?

djmitche: For the error and JSON, I think we can assert that some part of those appear in the content using a regular expression.

I don't think `this` would have worked, actually, since this isn't a class.  But my knowledge of how JS classes work is a little spotty.

hammad13060: I think we can check for following phrases
```The hooks service was unable to create a task for hook ${hook.hookGroupId}/${hook.hookId}```

```The error was:```
```Details:```

Comment (djmitche) [inline src/scheduler.js]: Sounds good.

File: test/scheduler_test.js [hunk 9]
```diff
@@ -178,5 +182,22 @@ suite('Scheduler', function() {
...
       assume(updatedHook.nextScheduledDate).is.not.equal(oldScheduledDate);
     });
+
+    test('on error, notify is used with correct options', async () => {
+      creator.shouldFail = true;
+      hook.metadata.emailOnError = true;
+      await scheduler.handleHook(hook);
+
+      assume(scheduler.notify.credentials).exists();
+      assume(scheduler.notify.authorizedScopes).exists();
```
hammad13060: wondering if we should check for credentials and authorizedScopes because I think tc-notify api doesn't expose any such thing. These checks might be misleading. What do you say ?

djmitche: I agree, and your reasoning is exactly right.

PR #51, opened 2017-12-19
Passing two new options to app

yash-iiith:


yash-iiith:
@djmitche

djmitche:
Also: a good commit message might be:

```
Enable a docs link at the root URL

Taskcluster-lib-app now supports generating a link to the documentation,
and requires the `docs` object to do so, so pass that to its constructor.
```

petemoore:
@yash-iiith Was this intentionally closed, or did you mean to merge it? Many thanks!

djmitche:
It was fixed in #57

PR #52, opened 2017-12-22
Enable a docs link at the root URL

yash-iiith:
Taskcluster-lib-app now supports generating a link to the documentation, and requires the `docs` object to do so, so pass that to its constructor.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -103,7 +103,16 @@ var load = loader({
   server: {
     requires: ['cfg', 'router', 'docs'],
     setup: ({cfg, router, docs}) => {
-      let hooksApp = app(cfg.server);
+      let hooksApp = app({
+        port:                   cfg.server.port,
+        env:                    cfg.server.env,
+        forceSSL:               cfg.server.forceSSL,
+        trustProxy:             cfg.server.trustProxy,
+        contentSecurityPolicy:  cfg.server.contentSecurityPolicy,
+        robotstxt:              cfg.server.robotstxt,
+        rootDocsRedirect:       true, 
+        docs,
+      });
       hooksApp.use('/v1', router);
       return hooksApp.createServer();
     },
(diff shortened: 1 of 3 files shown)


yash-iiith:
@djmitche

jonasfj:
Looks good to me.. .but I'll leave it to bstack or dustin to merge... we're all on holidays now...

src/main.js
```diff
@@ -103,7 +103,16 @@ var load = loader({
   server: {
     requires: ['cfg', 'router', 'docs'],
     setup: ({cfg, router, docs}) => {
-      let hooksApp = app(cfg.server);
+      let hooksApp = app({
+        port:                   cfg.server.port,
```
djmitche: It looks like publicUrl is missing here?

src/main.js
```diff
@@ -103,7 +103,16 @@ var load = loader({
...
+        env:                    cfg.server.env,
+        forceSSL:               cfg.server.forceSSL,
+        trustProxy:             cfg.server.trustProxy,
+        contentSecurityPolicy:  cfg.server.contentSecurityPolicy,
+        robotstxt:              cfg.server.robotstxt,
```
djmitche: This should be `robotsTxt` (capital t).  Or better yet, it can be omitted here since it is not in `config.yml`, and defaults to true.  Same for `contentSecurityPolicy`.

djmitche:
Whoops, I meant 'Request Changes', but you get the idea.

I will be around periodically over the next week, but we are officially on holiday until January 3 so responses might be a little slow.

src/main.js
yash-iiith: but publicUrl is not being used in tc-lib-app. Right?

src/main.js
yash-iiith: Okay .

yash-iiith:
@djmitche

djmitche:
https://hooks.taskcluster.net/

---

You're lost, here's the way out.
[Refer to the documentation](https://hooks.taskcluster.net/undefined)

---

Almost there, but not quite (note the link destination)

PR #54, opened 2018-01-03
a brief description about use of json-e in context

alexandrasp:


src/v1.js
```diff
@@ -24,6 +24,11 @@ var api = new API({
     'https://www.npmjs.com/package/cron-parser.  For example:',
     ' * `[\'0 0 1 * * *\']` -- daily at 1:00 UTC',
     ' * `[\'0 0 9,21 * * 1-5\', \'0 0 12 * * 0,6\']` -- weekdays at 9:00 and 21:00 UTC, weekends at noon',
+    '',
+    'Hooks can change with json-e context fields when you re-trigger a hook',
```
djmitche: This is a bit *too* brief!  How about this:

---

Hooks can be *parameterized* using [JSON-e](https://github.com/taskcluster/json-e).  The task definition in the hook is used as the JSON-e template, and the parameters are supplied as the JSON-e context.  The result of rendering this template and context is used as the task definition.

Currently context can only be provided with the `triggerHook` method.

alexandrasp: Ok Dustin, I will try not be too brief and write documentation for humans..

PR #55, opened 2018-01-03
Use the right property name to detect real credentials

djmitche:
I assume this was an error on my part in trying to make the tests
pending when there are no credentials.


alexandrasp:
Ok Dustin, make more sense now. :) 
The code sounds doesn't know what is _hasTcCredentials_, but _haveRealCredentials_.

PR #57, opened 2018-01-03
Changed rootDocsRedirect to rootDocsLink

yash-iiith:
Changed rootDocsRedirect to rootDocsLink as it is being used as rootDocsLink in tc-lib-app


yash-iiith:
@djmitche

PR #64, opened 2018-01-23
correct message for multiple schema validation errors

alexandrasp:


test/api_test.js
```diff
@@ -263,8 +263,15 @@ suite('API', function() {
...
-        (err) => { assume(err.statusCode).equals(400); });
+        triggeredBy: 'triggerHook'}).then(() => { throw new Error('Expected an error'); },
+        (err) => { debug('Got expected error: %s', err); });
+    });
+
+    test('checking more than one schema validation error', async () => {
+      await helper.hooks.createHook('foo', 'bar', hookWithTriggerSchema);
+      await helper.hooks.triggerHook('foo', 'bar', {location: 28, otherVariable: 'twelve',
+        triggeredBy: 'triggerHook'}).then(() => { throw new Error('Expected an error'); },
+        (err) => { debug('Got expected error: %s', err); });
```
djmitche: This should probably check that the error was as expected!  Something like `/data.location should be string/.test(err.toString())` and a similar check for the second error.  This context only seems to generate one error -- I think because this patch removes `{format: 'full', verbose: true, allErrors: true}` from the Ajv constructor.

src/v1.js
```diff
@@ -373,7 +373,7 @@ api.declare({
   var payload = req.body;
   var hook = await this.Hook.load({hookGroupId, hookId}, true);
   var error = null;
-  const ajv = new Ajv({format: 'full', verbose: true, allErrors: true});
+  const ajv = new Ajv();
```
djmitche: Why this change?

test/api_test.js
alexandrasp: yes, you are right! 
all errors -> makes the code return all errors, not only the first
format -> validate formats (data, time, etc)

src/v1.js
alexandrasp: because I was testing, but verbose and allErrors are really important to "understand" what are doing the schemas fail, and of course format in cases that data, or hours are being used.

PR #66, opened 2018-01-27
triggerSchema default in createHook

alexandrasp:
@imbstack I put an empty triggerSchema by default in hookDef that doesn't have one and it seems to solve the problem. What do you think about?

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -230,7 +230,6 @@ api.declare({
         '{{message}} in {{schedule}}', {message: err.message, schedule});
     }
   });
-
   // Try to create a Hook entity
   try {
     var hook = await this.Hook.create(
--- a/schemas/create-hook-request.yml
+++ b/schemas/create-hook-request.yml
@@ -31,7 +31,8 @@ properties:
   deadline:               {$const: deadline}
   task:                   {$ref: "http://schemas.taskcluster.net/hooks/v1/task-template.json"}
   triggerSchema:
-    type:                 object
+    type:                 'object'          
+    default:              {type: "object", additionalProperties: false}
 additionalProperties:     false
 required:
   - metadata


src/v1.js
```diff
@@ -233,6 +233,7 @@ api.declare({
 
   // Try to create a Hook entity
   try {
+
```
djmitche: Careful about adding unnecessary empty lines like this.  They don't hurt anything but they take up space on our screens! :)

src/v1.js
```diff
@@ -217,7 +217,7 @@ api.declare({
   var hookId    = req.params.hookId;
   var hookDef   = req.body;
 
-  hookDef = _.defaults({hookGroupId, hookId}, hookDef);
+  hookDef = _.defaults({hookGroupId, hookId}, hookDef, {triggerSchema:{}});
```
djmitche: `{}` will match any object, but by default we want a schema that only matches an empty object.  I think that is `{"type": "object", "additionalProperties": false}`.

Applying the defaults here works fine, but we can also apply them at https://github.com/taskcluster/taskcluster-hooks/blob/c545be7b904b1a014a294ccbe336536299fea6b8/schemas/create-hook-request.yml#L33-L34 which would make the default visible in the documentation.

I believe that if a `default` is specified in the schema, then the change here in `v1.js` won't be required.  But the tests will check that I'm right :)

alexandrasp: @djmitche, the tests fail again :/ 
I wrote a test to check if a hook created without any triggerSchema could receive this by default, but the test doesn't pass. I've put : `{"type": "object", "additionalProperties": false}` in `create-hook-request.yml`.

Now, I don't know if we can put this using `_.defaults` in case that doesn't exist a previous definition, and if the test is checking exactly this (https://github.com/taskcluster/taskcluster-hooks/pull/66/files#diff-8eda7840a1b7af5ed4deca58f094f20dR29).

src/v1.js
alexandrasp: it was removed. :)

src/v1.js
djmitche: With this change to this PR:

```patch
diff --git a/schemas/create-hook-request.yml b/schemas/create-hook-request.yml
index 1461c2d..98fdbf4 100644
--- a/schemas/create-hook-request.yml
+++ b/schemas/create-hook-request.yml
@@ -27,13 +27,14 @@ properties:
           Note that tasks may not be created at exactly the time specified.
     uniqueItems: true
     default: []
   expires:                {$const: expires}
   deadline:               {$const: deadline}
   task:                   {$ref: "http://schemas.taskcluster.net/hooks/v1/task-template.json"}
   triggerSchema:
     type:                 object
+    default:              {"type": "object", "additionalProperties": false}
 additionalProperties:     false
 required:
   - metadata
   - task
   - deadline
diff --git a/src/v1.js b/src/v1.js
index 826eec2..4fc5369 100644
--- a/src/v1.js
+++ b/src/v1.js
@@ -212,17 +212,17 @@ api.declare({
     'create the task.  That role must satisfy task.scopes as well as the',
     'necessary scopes to add the task to the queue.',,
   ].join('\n'),
 }, async function(req, res) {
   var hookGroupId = req.params.hookGroupId;
   var hookDef   = req.body;
 
-  hookDef = _.defaults({hookGroupId, hookId}, hookDef, {triggerSchema:{}});
+  hookDef = _.defaults({hookGroupId, hookId}, hookDef);
 
   await req.authorize({hookGroupId, hookId});
 
   // Validate cron-parser expressions
   _.forEach(hookDef.schedule, function(schedule) {
     try {
       parser.parseExpression(schedule);
     } catch (err) {
```

the tests pass..

alexandrasp: additionalProperties false?

djmitche: In this case, yes -- we want a schema that *only* allows `{}`, so it has zero properties and allows no additional properties.

djmitche:
I successfully created and updated a hook :)

PR #67, opened 2018-01-29
Document use of JSON-e, make it more consistent

djmitche:
This is nicely broken up into commits, I hope.  Note that 8200a68 is a bugfix: triggering hooks with a payload currently uses the wrong context!

The rest are minor improvements, either improving consistency of terminology ("fire" vs. "trigger", "payload" vs. "context") or defining and adding JSON-e contexts to the firing mechanisms that didn't already have them (`schedule` and `triggerHookWithToken`).

### DIFF
--- a/docs/firing-hooks.md
+++ b/docs/firing-hooks.md
@@ -40,7 +40,7 @@ against the hook's `triggerSchema`, and supplied in the JSON-e context as
 ```
 {
     firedBy: "triggerHook",
-    context: {..}               // API call payload
+    payload: {..}               // API call payload
 }
 ```
 
--- a/docs/firing-hooks.md
+++ b/docs/firing-hooks.md
@@ -48,7 +48,19 @@ The schema validation also applies any default values specified in the schema.
 
 ### Webhooks
 
-Not yet implemented (context is `{}`)
+Webhooks are invoked with the `triggerHookWithToken` API method. This method is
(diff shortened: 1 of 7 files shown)


docs/firing-hooks.md
```diff
@@ -0,0 +1,69 @@
...
+
+### TriggerHook
+
+Calls to the `triggerHook` method include a payload. The payload is validated
+against the hook's `triggerSchema`, and supplied in the JSON-e context as
+`context`:
+
+```
+{
+    firedBy: "triggerHook",
```
alexandrasp: all this call the fire function, so the most proper name is firedBy! Good.

src/v1.js
```diff
@@ -349,6 +347,10 @@ api.declare({
   stability:    'experimental',
   description: [
     'This endpoint will trigger the creation of a task from a hook definition.',
+    '',
+    'The HTTP payload must match the hook\s `triggerSchema`.  If it does, it is',
+    'provided as the `payload` property of the JSON-e context used to render the',
```
alexandrasp: payload? Ok, make sense if you are thinking in get some info, but is it a JSON-e context, right?

djmitche: It will be a sub-object of the JSON-e context:
```
{
  firedBy: "triggerHook",
  payload: payload,
}
```

alexandrasp: Sure.

PR #68, opened 2018-01-30
Added a description for triggerSchema and provided a schema

petemoore:
I wanted to add a description and schema for `triggerSchema` property, but I'm not sure I've defined it very well or used the correct terminology.

I'm also assuming that triggerSchema should always be a valid json schema - if that's not the case, let me know! :-)

Thanks!

### DIFF
--- a/schemas/create-hook-request.yml
+++ b/schemas/create-hook-request.yml
@@ -21,18 +21,23 @@ properties:
       title: "Cron Pattern"
       type: string
       description: |
-          Cron-like specification for when tasks should be created.  The pattern is
-          parsed in a UTC context.
-          See [cron-parser on npm](https://www.npmjs.com/package/cron-parser).
-          Note that tasks may not be created at exactly the time specified.
+        Cron-like specification for when tasks should be created.  The pattern is
+        parsed in a UTC context.
+        See [cron-parser on npm](https://www.npmjs.com/package/cron-parser).
+        Note that tasks may not be created at exactly the time specified.
     uniqueItems: true
     default: []
   expires:                {$const: expires}
   deadline:               {$const: deadline}
   task:                   {$ref: "http://schemas.taskcluster.net/hooks/v1/task-template.json"}
   triggerSchema:
+    description: |
+      The hook definition is a json-e template which is combined with a trigger
+      payload to form a task definition. The `triggerSchema` property defines a
+      json schema to validate the trigger payload against.
     type:                 'object'          
     default:              {type: "object", additionalProperties: false}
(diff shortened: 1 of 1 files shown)


schemas/create-hook-request.yml
```diff
@@ -21,18 +21,23 @@ properties:
...
     uniqueItems: true
     default: []
   expires:                {$const: expires}
   deadline:               {$const: deadline}
   task:                   {$ref: "http://schemas.taskcluster.net/hooks/v1/task-template.json"}
   triggerSchema:
+    description: |
+      The hook definition is a json-e template which is combined with a trigger
+      payload to form a task definition. The `triggerSchema` property defines a
+      json schema to validate the trigger payload against.
```
djmitche: Please add a link to the firing-hooks doc (see the API description at the top of this file) for the details of the format of this context, since the payload and the context aren't exactly the same, and the difference can be confusing :)

schemas/create-hook-request.yml
```diff
@@ -21,18 +21,23 @@ properties:
...
     type:                 'object'          
     default:              {type: "object", additionalProperties: false}
+    $ref:                 'http://json-schema.org/draft-06/schema#'
```
djmitche: Ah, I didn't realize it was this easy!

djmitche (changes requested):

alexandrasp (approved):

schemas/create-hook-request.yml
alexandrasp: @djmitche said the most important point here, payload and the context aren't exactly the same and is important avoid any confusion with this, I had a lot..

alexandrasp (changes requested):

schemas/create-hook-request.yml
petemoore: Unfortunately it won't help enormously, because in http://json-schema.org/draft-06/schema the top level `additionalProperties` is not set to false. In other words, a draft 6 json schema is valid even if it contains unknown terms (such as misspelling `required` as `require` or even `requiredProperties` etc). So we might want to add extra constraints like this and creating a custom schema that is more restrictive, but we may end up doing this as part of [bug 1434203](https://bugzilla.mozilla.org/show_bug.cgi?id=1431750) where we are already intending to impose restrictions against our schemas to prevent us making foolish mistakes. This would mean updating this schema reference to our more formal one, once we've written it.

djmitche:
bitrotten..

djmitche:
(@petemoore please feel free to reopen on the monorepo)

PR #70, opened 2018-02-05
also update the triggerSchema when editing

djmitche:


djmitche:
Yes, it will be necessary in your PR too :)

PR #73, opened 2018-02-07
removing bindings, pulseExchanges will be use to pulsed messages

alexandrasp:


djmitche:
Ah, I didn't see that you'd updated this.

alexandrasp:
sure, @djmitche

PR #76, opened 2018-02-18
handle invalid schemas - bug 1430902

alexandrasp:


test/api_test.js
```diff
@@ -314,6 +297,52 @@ suite('API', function() {
...
+        otherVariable: 'twelve',
+        foo: 'triggerHook',
+      }).then(() => { throw new Error('Expected an error'); },
+        (err) => { debug('Got expected error: %s', err); });
+    });
+
+    test('handle an invalid schema', async () => {
+      nHookDef = _.defaults({
+        triggerSchema: {
+          type: 'beer',
```
djmitche: I feel like that *should* be valid... ;)

djmitche:
@alexandrasp did you want to finish this up?

alexandrasp:
@djmitche yes :)

alexandrasp:
@djmitche in triggerHook make sense use directly context.payload to validate?

djmitche:
There are two validations we're concerned with here:

 * Is trigerSchema a valid JSON schema?  
 * Is context.payload a valid trigger payload?  That is, does it satisfy triggerSchema?

The second, we already do correctly.  This bug is to do the first.

There is a schema for schemas (called a meta-schema), but my thinking was to just try calling `ajv.compile(triggerSchema)` any time we get a new triggerSchema.  If that fails (throws an error), return a 400 InputError.  Now that I write it out, we should probably use the meta-schema.  I did some experimentation on the command line:

```
> (v8.10.0) dustin@lamport ~/p/taskcluster-hooks [bug1446966*] $ curl http://json-schema.org/draft-06/schema > json-schema.json
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4445  100  4445    0     0  15041      0 --:--:-- --:--:-- --:--:-- 15067
(v8.10.0) dustin@lamport ~/p/taskcluster-hooks [bug1446966*] $ node
> const Ajv = require('ajv');
undefined
> const ajv = new Ajv({format: 'full', verbose: true, allErrors: true});
undefined
> ajv.addSchema(JSON.parse(fs.readFileSync('json-schema.json')))
Ajv {
.....
}
> validate = ajv.getSchema("http://json-schema.org/draft-06/schema")
{ [Function: validate]
....
}
> validate({type: "string"})
true
> validate({type: 10})
false
```

@jonasfj does tc-lib-validate make this easier?  Do you have any other advice?  Should we keep a copy of the schema in tc-hook's `schemas/` directory?

jonasfj:
I would advice using a custom meta-schema... and a simple one at that...
There is a lot of features in JSON schema making them all available easily makes the API implementation dependent on ajv.

The default meta schema allows all sorts of additional properties, it is hence not forward compatible with newer version of JSON schema where such properties may have semantics.

>  Should we keep a copy of the schema in tc-hook's schemas/ directory?
Yes, keeping the meta-schema in `schemas/` makes sense... You can probably `$ref` it from the API.

djmitche:
I'm a bit confused by that comment.  What would a "custom" meta-schema mean?  Does that mean that hooks has its own special version of JSON schema?  That seems way too detailed..

jonasfj:
Note, I don't think using the official meta-schema is necessarily a bad idea.
But it's extremely flexible, if we want to encourage people to make a sane design for schemas we could something rather strict like:

```yml
anyOf:
  - title: 'string'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'string'}
      minLength: {type: integer, minimum: 0}
      maxLength: {type: integer, minimum: 0}
      pattern: {type: string, format: regex}
      format: {enum: [date-time, email, hostname, uri]}
      enum: {type: array, items: {type: string}, minItems: 1, uniqueItems: true}
    required: [title, description, type, maxLength]
    additionalProperties: false
  - title: 'integer'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'integer'}
      minimum: {type: integer}
      maximum: {type: integer}
      exclusiveMinimum: {type: integer}
      exclusiveMaximum: {type: integer}
      enum: {type: array, items: {type: integer}, minItems: 1, uniqueItems: true}
    required: [title, description, type]
    additionalProperties: false
  - title: 'number'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'number'}
      minimum: {type: number}
      maximum: {type: number}
      exclusiveMinimum: {type: number}
      exclusiveMaximum: {type: number}
      enum: {type: array, items: {type: number}, minItems: 1, uniqueItems: true}
    required: [title, description, type]
    additionalProperties: false
  - title: 'boolean'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'boolean'}
    required: [title, description, type]
    additionalProperties: false
  - title: 'object'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'object'}
      properties: {type: object, additionalProperties: {$ref: '#'}}
      required: {type: array, items: {type: string}, uniqueItems: true}
      additionalProperties: {const: false}
    required: [title, description, type, properties, required, additionalProperties]
    additionalProperties: false
  - title: 'map'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'object'}
      additionalProperties: {$ref: '#'}
    required: [title, description, type, additionalProperties]
    additionalProperties: false
  - title: 'array'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'array'}
      items: {$ref: '#'}
      minItems: {type: integer, minimum: 0}
      maxItems: {type: integer, minimum: 0}
      uniqueItems: {type: boolean}
    required: [title, description, type, items]
    additionalProperties: false
  - title: 'json'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      type: {const: 'object'}
    required: [title, description, type]
    additionalProperties: false
  - title: 'anyOf'
    type: object
    properties:
      title: {type: string}
      description: {type: string}
      anyOf: {$ref: '#'}
    required: [anyOf]
    additionalProperties: false
```

This limits what features can be used, and forces specification of properties like `title` and `description` as well as properties that are often accidentally missed like `additionalProperties`. It also forces schemas for objects to be either an object where all properties are specified, or a map where properties all map to thing matching the same schema. This is super opinionated, just a suggestion.

Whether to be as strict I proposed above is a personal choice, we can always allow more features.  But removing them is hard.

-----
It's also fair to just disregard this non-sense I'm writing and allow the full meta schema... The problem is that meta-schema doesn't forbid properties that have no meaning at this point. So if people add those and we upgrade ajv to a newer version those properties might start meaning something and break hooks deployed.

Another option is to make a stricter form of the meta-schema, rather than a hyper opinionated thing like I did... example could be something like:
https://gist.github.com/jonasfj/53729bc15e5b99948ffb7af86edd8d65#file-schema-js-L168

djmitche:
I like the very last option -- taking the existing meta-schema and adding `additionalProperties: false` to it.

I worry that if we're introducing our own JSON schema variant, we're going to get into a world of hurt where we get something wrong or omit a feature and want to add it, then we have slightly different schemas acceptd here and there, and so on.  As an example, I'm pretty opposed to requiring title and description everywhere -- often they're obvious (e.g., when the title and the property name are the same).

So, @alexandrasp -- I think the right way to proceed is to take the schema in Jonas's gist and copy it into `schemas/` (translating it to YAML in the process).  Its `$id` field should also be omitted, as it isn't that schema anymore.  Then, use $ref to refer to that schema is hook, createHook, and updateHook and anywhere else we take or return a hook definition.  That should be enough -- API validation will then catch any invalid triggerSchemas.  We should add a test just to check :)

alexandrasp:
working on it

djmitche:
OK, I updated this to just verify that the schema is valid, and rebased and fixed conflicts.  I also removed some of the additional validation added in triggerHook, as that wasn't part of this bug.  @jonasfj was our JSON-schema expert, so without him here I don't want to get into custom meta-schemas, etc.

src/v1.js
```diff
@@ -204,6 +205,11 @@ builder.declare({
     }
   }
 
+  //handle an invalid schema
+  if (!ajv.validateSchema(hookDef.triggerSchema)) {
+    return res.reportError('InputError', 'Invalid Schema', {});
```
imbstack: Should we maybe specifically reject schemas that aren't of the correct version? i.e. draft-6 or something?

djmitche: I think the idea here is to catch most schema errors during the API call. If it turns out that the schema has some other issue, that will appear when triggered. So, I don't think we need to try to do anything fancy (and bug-prone) here.

PR #77, opened 2018-02-27
Remove task-specific schema from hook.task

kritisingh1:
This is in reference to Bug 1438019 which requires updating the current schemas in accordance with JSON-e so that hook.task has `type: object`.


kritisingh1:
@djmitche the test that calls createTask still needs to be added, right? Thanks :)

djmitche:
Oh right!  Please make a new PR for that.  Silly me!!

kritisingh1:
Sure. Thanks a lot!

PR #78, opened 2018-02-27
Bug 1437911 - upgrade azure-entities

djmitche:
Just an upgrade of a dependency to get better error handling.


djmitche:
I just made the update to azure-entities a few days ago.

PR #80, opened 2018-02-28
Updated the tests for the task property 

kritisingh1:


djmitche:
It's failing like this now:

```
  4) TaskCreator
       adds a taskId if one is not specified:
     Error: 
Schema Validation Failed!
Rejecting Schema: http://schemas.taskcluster.net/queue/v1/create-task-request.json#
Errors:
  * data should NOT have additional properties
  * data should have required property 'provisionerId'
  * data should have required property 'workerType'
  * data should have required property 'payload'
  * data should have required property 'metadata'
----
method:     createTask
errorCode:  InputValidationError
statusCode: 400
time:       2018-02-28T22:22:21.263Z
```

I think that the problem is, you've emptied out the task definitions, so the taskcreator is calling `createTask({})` (with an empty task).

Let's try something like this, which is JSON-e but should produce a valid task definition when rendered:
```json
{
  $if: "true",
  then: {
    provisionerId: 'no-provisioner',                                                                                                                                                     
    workerType: 'test-worker',
    metadata: {
      name: 'test task',
      description: 'task created by tc-hooks tests',
      owner: 'taskcluster@mozilla.com',
      source: 'http://taskcluster.net',
    },  
    payload: {}, 
  }
}
```

(basically all I did was take the old value and embed it in `{$if: 'true', then: <old value>}`).

kritisingh1:
@djmitche do the tests successfully pass now, with the credentials I mean?

djmitche:
Sorry, this is super-annoying for you, I'm sure.  We are about to make some fixes to how people login to taskcluster which would make it easier to give you credentials so you can run these tests yourself.  But in the interest of you not having to do another round of "programming blind" and then asking for a test run, I fixed up the last failures.  The remaining problem was that a lot of tests modified the hook by setting things like `hook.task.extra`.  With the change to JSON-e, that makes an invalid construct `{$if: .., then: .., extra: ..}`.  So I changed it to `hook.task.then.extra`.

djmitche:
Thanks for your patience :)

kritisingh1:
Oh so the `then` was the catch. JSON-e thing is quite helpful in the sense it makes code easier to understand as it treats each element as an object. I will take care of it from next time. I am not at all annoyed for sure :), in fact, it is an amazing thing to learn. And I should thank you for reviewing every 'programming blind' of newbies like me with patience. Thanks a lot :)

PR #81, opened 2018-03-01
Added test for created, deadline and expires fields

kritisingh1:
This is in reference to Bug 1437552 : Remove deprecated `deadline` and `expires` fields. 
As a first step I have tried to create test that creates a hook with created, deadline and expires fields.


djmitche:
I think this was superseded by #85

PR #84, opened 2018-03-05
Bug 1442717 - Allow `hookId` and `hookGroupId` in createHook / updateHook

kanikasaini:
I see that createHook API method already uses _.defaults to set the default value for hookGroupId and hookId if they are not present. If they are present, they are checked using req.authorize(). 

In updateHook API method, I have added the functionality to set default values to these properties if they are not present, and they are being checked by req.authorize() if they are already there.

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -215,6 +215,12 @@ api.declare({
   var hookId    = req.params.hookId;
   var hookDef   = req.body;
 
+  if (req.body.hookGroupId !== null && req.body.hookId !== null) {
+    if (hookGroupId !== req.body.hookGroupId || hookId !== req.body.hookId) {
+      return res.reportError('InputError', 'Hook Ids do not match', {});
+    }
+  }
+
   hookDef = _.defaults({hookGroupId, hookId}, hookDef);
 
   await req.authorize({hookGroupId, hookId});
@@ -277,6 +283,12 @@ api.declare({
   var hookGroupId = req.params.hookGroupId;
   var hookId = req.params.hookId;
   var hookDef = req.body;
+
+  if (req.body.hookGroupId !== null && req.body.hookId !== null) {
+    if (hookGroupId !== req.body.hookGroupId || hookId !== req.body.hookId) {
+      return res.reportError('InputError', 'Hook Ids do not match', {});
+    }
+  }
 
   hookDef = _.defaults({hookGroupId, hookId}, hookDef);
 
--- a/src/v1.js
+++ b/src/v1.js
@@ -215,13 +215,16 @@ api.declare({
   var hookId    = req.params.hookId;
   var hookDef   = req.body;
 
-  if (req.body.hookGroupId !== null && req.body.hookId !== null) {
-    if (hookGroupId !== req.body.hookGroupId || hookId !== req.body.hookId) {
-      return res.reportError('InputError', 'Hook Ids do not match', {});
-    }
+  if (req.body.hookGroupId && hookGroupId !== req.body.hookGroupId) {
+    return res.reportError('InputError', 'Hook Group Ids do not match', {});
   }
 
-  hookDef = _.defaults({hookGroupId, hookId}, hookDef);
+  if (req.body.hookId && hookId !== req.body.hookId) {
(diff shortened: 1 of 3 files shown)


src/v1.js
```diff
@@ -215,6 +215,12 @@ api.declare({
   var hookId    = req.params.hookId;
   var hookDef   = req.body;
 
+  if (req.body.hookGroupId !== null && req.body.hookId !== null) {
```
djmitche: If not given, these will not be `null`, but undefined.  You can check for that with `if (req.body.hookGroupId)`.  You'll need to check the id's separately, too -- as written, if I pass just `hookGroupId`, this conditional will be false and the code to check the id's match will never execute!

src/v1.js
```diff
@@ -215,6 +215,12 @@ api.declare({
...
+    if (hookGroupId !== req.body.hookGroupId || hookId !== req.body.hookId) {
+      return res.reportError('InputError', 'Hook Ids do not match', {});
+    }
+  }
+
   hookDef = _.defaults({hookGroupId, hookId}, hookDef);
```
djmitche: This line is a bit security-sensitive, since it's setting the hook's actual group and id.  And `_.defaults` has behavior that's not obvious.  Imagine this scenario:

* I am allowed to create hooks with hookGroupId "garbage" and any hookId.
* hooks in the "firefox" group have special access to important things like shipping a new version of Firefox to 500 million people
* I call `createHook('garbage', 'dustins-hook', {hookGroupId: 'firefox, task: ..})

As written, what happens?

First, the access-control check is performed against the parameters, "garbage"/"dustins-hook".  I have permission, so that check passes.  Then:
```javascript
if (req.body.hookGroupId !== undefined && req.body.hookId !== undefined) { 
```
`req.body.hookId is undefined, so this is false.  So we skip the body of that conditional.
```javascript
hookDef = _.defaults({hookGroupId, hookId}, hookDef);
```
This will set `hookDef.hookId` to `'dustins-hook'`, but it will not change `hookDef.hookGroupId`, because it's already set -- to "firefox".  This is the non-obvious behavior I mentioned.

Result: I can create a hook in the "firefox" hookGroup, that ships a copy of Firefox with malware in it.  The end!

I don't really like that scenario!  So to protect against it, step 1 is to write the conditional correctly.  But it's best to have multiple lines of defense -- what if I didn't notice that in the review?  The other thing we can do is to *always* set the hookGroupId and hookId based on the parameters -- whether they are already set or not.   So:

```javascript
hookDef.hookGroupId = hookGroupId;
hookDef.hookId = hookId;
```
This will always be correct.  If we're overwriting a value that was already in hookDef, that's OK -- we just checked that it matched anyway.

src/v1.js
```diff
@@ -278,6 +284,14 @@ api.declare({
   var hookId = req.params.hookId;
   var hookDef = req.body;
 
+  if (req.body.hookGroupId !== null && req.body.hookId !== null) {
+    if (hookGroupId !== req.body.hookGroupId || hookId !== req.body.hookId) {
+      return res.reportError('InputError', 'Hook Ids do not match', {});
+    }
+  }
+
+  hookDef = _.defaults({hookGroupId, hookId}, hookDef);
```
djmitche: same thing here :)

kanikasaini:
@djmitche Thanks a lot! :)
I have updated the files. 
However, yarn test still gives an error.

djmitche:
> Also, I'm not sure what properties do 'foo' and 'bar' represent in this call in api_test.js
var r1 = await helper.hooks.createHook('foo', 'bar', hookWithTriggerSchema)

sorry, I missed this question.  Here, `foo` is a hookGroupId and `bar` is a hookId.

The `EADDRINUSE` bug that you're seeing is because something else on your system is already running on port 60401.  Most of the time, that's just another instance of the test scripts.  Finding that process (with `ps`) and killing it (with `kill`) can help, or if you want to take the fast way restarting the computer will do it too.

I ran the tests locally, and they all passed, so I'll have a look at the patch shorty.

kanikasaini:
@djmitche tests added and they run successfully :)

djmitche:
Deployed!

PR #89, opened 2018-03-26
Bug 1437552 - remove hook.deadline / hook.expires

djmitche:
..but keep the task-related fields of the same name.

* In `data.js`, this adds a new version of the table, dropping those two fields, but does not modify older versions
* In tests, this passes deadline/expires as task-level properties everywhere
* Removes deadline/expiration from constants, but adds them in `task-status.yml` as copied from taskcluster-queue's schemas
* In TaskCreator, no longer sets task.created/expires/deadline after rendering

@kritisingh1 what do you think?


test/taskcreator_test.js
```diff
@@ -10,10 +10,12 @@ suite('TaskCreator', function() {
   this.slow(500);
   helper.setup();
 
-  // these tests require real TaskCluster credentials (for the queue insert)
-  if (!helper.haveRealCredentials && !process.env.TASK_ID) {
-    this.pending = true;
-  }
+  setup(function() {
```
djmitche: Putting this in a `setup` block makes it run after helper's `setupSuite` has run (which is what sets `helper.haveRealCredentials`). Without this, these tests never run! /cc @catlee fiennyangeln

kritisingh1:
I think that I missed this `Removes deadline/expiration from constants, but adds them in task-status.yml as copied from taskcluster-queue's schemas` in the previous work. @djmitche could you please explain its relevance a bit more? Like it prevents breaking some code or is to maintain consistency? Thank you :)

djmitche:
Yes, without that there were errors because task-status.yml has `{$constant: 'deadline'}` but that constant wasn't defined.

imbstack:
Is the plan here to turn off the travis stuff?

djmitche:
I think so, yes.

djmitche:
Rebased against the new master.  Note that this really just stops *returning* these properties -- it's been invalid to send them for quite some time now.

PR #92, opened 2018-04-14
Bug 1453724, 1452000 - upgrade tc-lib-monitor to get a new raven, enable option

djmitche:


djmitche:
Bumped to 5.3.0 for bug 1453725.

PR #93, opened 2018-04-18
JSON schema normalisation and cleanup

petemoore:
Hope you'll like this. 😉


jonasfj:
Not sure if the ref to schema will work in production

PR #96, opened 2018-04-23
Bug 1455697 - provide newTaskId in context, too

djmitche:


djmitche:
Sorry @jonasfj I'm going to need this soon :)

docs/firing-hooks.md
```diff
@@ -17,9 +17,10 @@ submits it to the Queue service via `Queue.createTask`.
 A hook definition's `task` property is, in fact, a [JSON-e](https://taskcluster.github.io/json-e/) template.
 When a hook is fired, that template is rendered and the result is submitted to `Queue.createTask`.
 
-The context for that rendering is an object with property `firedBy`, giving
-the action that led to the hook firing. The other properties of the object vary
-depending on this property.
+The context for that rendering is an object with property `firedBy`, giving the
+action that led to the hook firing; and `newTaskId` giving the taskId of the
```
jonasfj: Can we call it `taskId` or `ownTaskId`

jonasfj: Also what about `taskGroupId`

docs/firing-hooks.md
```diff
@@ -17,9 +17,10 @@ submits it to the Queue service via `Queue.createTask`.
...
+task being created. The other properties of the object vary depending on the
+`firedBy` property.
```
jonasfj: Please document and ensure that properties in the context can overwrite the built-in context properties... This way we can add more built-in properties without breaking compatibility..

docs/firing-hooks.md
djmitche: I'll rename it.  The taskGroupId = taskId so there's no need to specify it separately.  I'll add that to the doc, though.

PR #99, opened 2018-05-01
Bug 1446966 - try this one simple trick to improve your testing quality!

djmitche:
Similar to https://github.com/taskcluster/taskcluster-secrets/pull/38 but this app is a bit more complicated. For one thing, it uses the new `fake` support in taskcluster-client to fake out the notify service (which is otherwise not observable, so there's no "real" version)


djmitche:
woo, green on the first try this time :)

PR #102, opened 2018-05-21
Bug 1463243 - better error messages on InsufficientScopes errors

djmitche:
This is a very common error, so worth optimizing..


djmitche:
'uhoh' is handy because when you see it in a failed test, you know it's not a real error :)

PR #74, opened 2018-02-13
[WIP] Trigger pulse messages

djmitche:
PR to deploy to a Heroku test app

PR #108, opened 2018-09-16
 Bug 1225243 - Support triggering tasks in response to pulse messages

ydidwania:
I have so far updated the `create-hook-request` schema to include bindings and also defined exchanges to publish to from the api.
I have taken the `pulse-hook-message.yml` from alex's PR #74 (`hook-message.yml`)
I am not very sure on the schemas, so let me know what changes I have to make. Also I had trouble with deciding names, like `pulse-hook-message.yml` might not be obvious.
Things left to do are - 
* Publish pulse messages whenever we receive a request for a pulse based hook
* Create a new listeners component to listen to the exchanges and create hook-specific listener
* Tests
* Handle the case when a hook may respond to pulse messages created by its own task


ydidwania:
Hmm, the tests fail because of the missing pulse creds

schemas/v1/create-hook-request.yml
```diff
@@ -4,6 +4,27 @@ description: |
...
+      description: |
+        Exchange and RoutingKeyPattern for each binding
+      type:               object
+      properties:
+        exchange:
+          type:           string
+        routingKeyPattern:
+          type:           string
+        routingKeyReference:
+          type:           string
```
djmitche: Let's leave this off -- it's used by pulse consumers to disassemble the routing key into a data structure, but we won't be doing that here.

schemas/v1/pusle-hook-message.yml
```diff
@@ -0,0 +1,19 @@
+$schema:  http://json-schema.org/draft-06/schema#
+title:        "Hook Message"
```
djmitche: "Hook Changed Message"?  And maybe `pulse-hook-changed-message.yml`?

schemas/v1/pusle-hook-message.yml
```diff
@@ -0,0 +1,19 @@
...
+    description: |
+      `hookGroupId` of the hook that was changed
+    type:                 string
+  hookId:
+    description: |
+      `hookId` of the hook that was changed
+    type:                 string
+additionalProperties: true
+required:
+  - version
```
djmitche: I don't see this property..

src/exchanges.js
```diff
@@ -0,0 +1,109 @@
+let Exchanges = require('pulse-publisher');
+
+/** Declaration of exchanges used by hooks  */
+let exchanges = new Exchanges({
+  title:       'Exchnages to manage hooks',
```
djmitche: typo

src/exchanges.js
```diff
@@ -0,0 +1,109 @@
...
+    'Hooks are responsible for creating tasks',
+    'at specific times or in response to webhooks and API calls.',
+    'Using this exchange allows us to make hooks which',
+    'repsond to particular pulse messages.',
+    '',
+    'For each of the exchanges defined below,',
+    'there is a listener which creates ,updates or deletes',
+    'a hook. Basically it creates another listener,',
+    'with the neccessary bindings, which creates the task',
+    'for every message it receives.',
```
djmitche: https://github.com/taskcluster/taskcluster-auth/blob/cdfc489c7fcb463d021ce6907786e58a450cbe78/src/exchanges.js#L11-L18 might provide a good template for this.

src/exchanges.js
```diff
@@ -0,0 +1,109 @@
...
+  title:    'Hook Created Messages',
+  description: [
+    'Whenever the api receives a request to create a',
+    'pulse based hook, a message is posted to this exchange and',
+    'the receiver creates a listener with the bindings, to create a task',
+  ].join(''),
+  schema:   'pulse-hook-message.json#',
+  messageBuilder: commonMessageBuilder,
+  routingKey: buildCommonRoutingKey(),
+  routingKeyBuilder: commonRoutingKeyBuilder,
```
djmitche: confusing that we supply both a routingKey and routingKeyBuilder, but that seems to be the way pulse-publisher works!

src/exchanges.js
ydidwania: Thanks!
Is the hooks service available at `hooks.taskcluster.net` ?

djmitche: It is!

djmitche:
Looks like tests aren't passing -- quick fix?

ydidwania:
Looking into it

ydidwania:
I  backed out the `message-version` commit which were failing tests because I had not added that to `constants.yml`. The `version` field is passed in the `commonMessageBuilder` as of now.

The tests are failing now because

ydidwania:
I backed out the `version` field in `pulse-hooks-changed-message.yml`  and added env vars for pulse in `config.yml`
I have my pulse creds in `user-config.yml` locally and I am facing a different issue, the tests are passing but they are never ending. I don't get the `Done..` message at the end of it. [pastebin](https://pastebin.com/MRTkjTdh)

ydidwania:
I added the `withPulse` in `test/helpers.js` from [tc-queue](https://github.com/taskcluster/taskcluster-queue/blob/master/test/helper.js#L382), and I think there's some issue with that.

djmitche:
The non-termination issue sounds like something is forgetting to call `client.stop()` (to stop the tc-lib-pulse client).  Node sees that there are still setTimeout calls and TCP connections open and doesn't exit.

ydidwania:
@djmitche All tests locally pass for me. 
The failed test logs indicate something to do with a `bad decrypt` in `azure`, maybe  a wrong key. Can you check that?

djmitche:
That may be old data . I'll try deleting the table.

djmitche:
I deleted the HooksTestTable tables.  Azure takes a long time to delete tables, so I'll try to re-run the tests tomorrow (in 12h or so).

config.yml
```diff
@@ -42,6 +42,12 @@ defaults:
       secretAccessKey:  !env AWS_SECRET_ACCESS_KEY
       region:           us-west-2
       apiVersion:       2014-01-01
+  pulse:
+    username:       !env PULSE_USERNAME
+    password:       !env PULSE_PASSWORD
+    hostname:       !env PULSE_HOSTNAME
+    vhost:          !env PULSE_VHOST
+
```
djmitche: It isn't directly relevant to this PR, but you will probably want to use tc-lib-pulse, rather than PulseListener, when you start listening for arbitrary pulse messages.

ydidwania: yeah, I can use these [here](https://github.com/taskcluster/taskcluster-lib-pulse#client) as raw AMQP creds, right?

djmitche: Yep.  But let's not do that in this PR -- this one is close enough to land-able already :)

ydidwania:
What else do I need to send in the messages apart from `hookGroupId`, `hookId` and `task`?

djmitche:
I think that's all.  The other option would be to send the whole hook definition, but that seems superfluous.

djmitche:
I remove the `bindings` schema from the createHook schema, since it's not yet handled by the API methods.  I think that's the next part to work on, so feel free to just revert that commit as part of your work.

djmitche:
..removed it in https://github.com/taskcluster/taskcluster-hooks/commit/6aab59f16e4ed2bf33570d69ee3ec3adbe5df14e

djmitche:
screenshot_2018-09-21 pulse inspector taskcluster

djmitche:
:+1:

PR #109, opened 2018-09-25
[2] Bug 1225243 - Support triggering tasks in response to pulse messages

ydidwania:
I have completed the loader and client setup. 
Individual handlers for create, update and delete are still to be done


djmitche:
IIRC you were still working on this and not ready for review -- ping me when it is ready!

src/listeners.js
```diff
@@ -0,0 +1,145 @@
...
+        // `triggerSchema`,git  we can simply ignore the message.
+        const res = {
+          reply: function() {},
+          reportError: function() {},
+        };
+
+        await triggerHookCommon.call(this, {res, hook, payload, firedBy: 'pulseHook'});
+      });
+      for (let {exchange, routingKeyPattern} of oldBindings) {
+        await this.client.withChannel.unbindQueue(queueName, exchange, routingKeyPattern);
```
djmitche: I think this (and `bindQueue`) might fail in a case where we didn't update the Queues table correctly, so we should figure out how to handle that case correctly (probably with a try/catch, and verified with a test)

src/listeners.js
```diff
@@ -0,0 +1,145 @@
...
+      {},
+      {
+        limit: 1000,
+        handler:(queue) => Queues.push(queue),
+      }
+    );
+
+    let continuationToken = undefined;
+    
+    do {
```
djmitche: I think you could avoid having to deal with continuationToken here by putting the function you're passing to `forEach` into a `handler` parameter to `Hooks.scan`.

src/listeners.js
```diff
@@ -0,0 +1,145 @@
...
+        continuation: continuationToken,
+      });
+      continuationToken = allHooksData.continuation;
+      allHooksData.entries.forEach(hook => {
+        const {hookGroupId, hookId} = hook;
+        const queue = _.find(Queues, {hookGroupId, hookId});
+        if (queue) {
+          const index = this.listeners.findIndex(({_queueName}) => listener._queueName === queue.queueName);
+          if (index == -1) {
+            this.createListener(hook, queue.bindings);
```
djmitche: this is async and needs to be awaited (same below)

src/listeners.js
```diff
@@ -0,0 +1,145 @@
...
+          _.pull(Queues, queue);
+        } else {
+          this.createListener(hook, {});
+        }
+      });
+    } while (continuationToken);
+    
+    // Delete the queues now left in the Queues list.
+    Queues.forEach(async (queue) => {     
+      if (this.client.withChannel.checkQueue(queue.queueName)) {
```
djmitche: withChannel is a function that takes a callable, and calls it when the channel is available.

```js
await client.withChannel(async channel => { .. });
```

src/main.js
```diff
@@ -103,6 +105,53 @@ const load = loader({
...
+      return data.Queues.setup({
+        tableName: cfg.app.queuesTableName,
+        monitor: monitor.prefix('table.queues'),
+        credentials: sasCredentials({
+          accountId: cfg.azure.accountId,
+          tableName: cfg.app.queuesTableName,
+          rootUrl: cfg.taskcluster.rootUrl,
+          credentials: cfg.taskcluster.credentials,
+        }),
+        //cryptoKey: cfg.azure.cryptoKey,
```
djmitche: @Biboswan found a similar thing -- this table doesn't need any encrypted columns, so no need for encryption.

src/listeners.js
```diff
@@ -0,0 +1,145 @@
...
+        this.client.withChannel.deleteQueue(queue.queueName);
+      }
```
djmitche: I think you'll need to remove the listener from `this.listeners` here, too.

src/listeners.js
```diff
@@ -0,0 +1,145 @@
...
+        if (queue) {
+          const index = this.listeners.findIndex(({_queueName}) => listener._queueName === queue.queueName);
+          if (index == -1) {
+            this.createListener(hook, queue.bindings);
+          }
+          _.pull(Queues, queue);
+        } else {
+          this.createListener(hook, {});
+        }
+      });
```
djmitche: It doesn't look like anything is ever added to the Queues table?

src/main.js
ydidwania: Yes, that's why I commented it :)
Will remove it in the next commit

src/listeners.js
```diff
@@ -0,0 +1,145 @@
...
+      }, {
+        exchange: 'exchange/taskcluster-hooks/v1/hook-updated',
+        routingKeyPattern: '#',
+      }, {
+        exchange: 'exchange/taskcluster-hooks/v1/hook-deleted',
+        routingKeyPattern: '#',
+      }],
+      queueName: 'hookChanged',
+      maxLength : 50,
+      //expires : 1800000,
```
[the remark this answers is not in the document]
ydidwania: If I am deleting queues using `withChannel`, I shouldn't need to expire queues, right?

src/listeners.js
djmitche: This still stands.  And I didn't quite say that right -- you'll need to iterate over the listeners in `this.listeners` and call `listener.stop()` for each one.

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+          await this.client.withChannel(async channel => channel.deleteQueue(queue.queueName));
+        }
+      }
+      queue.remove();
+    });
+
+    // Will cancel all consumers
+    if (!this.client.isFakeClient) {
+      this.client.stop(); 
+    }
```
djmitche: The PulseListener class was given a running client, so it's a little awkward for that instance to then close the client -- what if something else was also using that client?  I admit, nothing else *is* using it right now, but that might not be the case forever.

So I think that you can omit calling `this.client.stop()`.

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  /** Create a new pulse consumer for a hook */
+  async createListener({hook, oldBindings}) {
+    this.hook = hook;
+    const client =  this.client;
+    const queueName = `${hook.hookGroupId}/${hook.hookId}`; // serves as unique id for every listener
+    const listener = await pulse.consume({
+      client,      
+      queueName: queueName,
```
djmitche: you can omit the second `queueName` here

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+    const queueName = `${hook.hookGroupId}/${hook.hookId}`; // serves as unique id for every listener
+    const listener = await pulse.consume({
+      client,      
+      queueName: queueName, 
+      maxLength : 50,
+    }, async ({payload}) => {
+      const hook = this.hook;
+      // Pass empty functions in res which are executed by `triggerHookCommon`
+      // This is more of a hack so that if the payload doesnt validate against
+      // `triggerSchema`,git  we can simply ignore the message.
```
djmitche: I think this comment no longer applies?

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+      await this.taskcreator.fire(hook, {firedBy:'pulseMessage', payload});
+    });
+    if (!this.client.isFakeClient) {
+      for (let {exchange, routingKeyPattern} of oldBindings) {
+        await this.client.withChannel(async channel => channel.unbindQueue(queueName, exchange, routingKeyPattern));
+      }
+      for (let {exchange, routingKeyPattern} of hook.bindings) {
+        await this.client.withChannel(async channel => {channel.bindQueue(queueName, exchange, routingKeyPattern);});
+      }
```
djmitche: For the case where the user adds a binding, this will in fact unbind all bindings, then re-bind them (along with the new one).  Some clever use of the `Set` class could probably avoid this, so that the only change to the queue in such a case is to call `bindQueue` for the new binding.

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+            debug('New queue..creating listener');
+            await this.createListener({hook, oldBindings: [{}]});
+            // Add to Queues table
+            debug('adding to Queues table');
+            await this.Queues.create({
+              hookGroupId,
+              hookId,
+              queueName: `${hookGroupId}/${hookId}`,
+              bindings: hook.bindings,
+            });
```
djmitche: This seems like something that `createListener` could handle, since it knows the queueName?

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+      continuationToken = allHooksData.continuation;
+      allHooksData.entries.forEach(async (hook) => {
+        if (hook.bindings != []) {
+          const {hookGroupId, hookId} = hook;
+          const queue = _.find(Queues, {hookGroupId, hookId});
+          if (queue) {
+            const index = this.listeners.findIndex(({_queueName}) => listener._queueName === queue.queueName);
+            if (index == -1) {
+              debug('Existing queue..creating listener');
+              await this.createListener({hook, oldBindings: queue.bindings});
```
djmitche: This call will call `pulse.consume()` again, for the same queueName, meaning that there are now two listeners defined for the same queue.  Perhaps it makes sense to split up creating and deleting listeners into two functions, and separately define a function to update bindings?  Then this case can just call the update-bindings function.

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+              queueName: `${hookGroupId}/${hookId}`,
+              bindings: hook.bindings,
+            });
+          }
+        }
+      });
+    } while (continuationToken);
+    
+    // Delete the queues now left in the Queues list.
+    Queues.forEach(async (queue) => {
```
djmitche: For parity with `createListener`, it probably makes sense to put this in a function named `removeQueue`.

src/listeners.js
```diff
@@ -0,0 +1,180 @@
...
+        }
+      }
+      // Delete from this.listeners
+      let removeIndex = this.listeners.findIndex(({_queueName}) => queue.queueName === _queueName);
+      if (removeIndex > -1) {
+        const listener = this.listeners[removeIndex];
+        listener.stop();
+        this.listeners.splice(removeIndex, 1);
+      }
+      queue.remove();
```
djmitche: This is an async operation and needs to be awaited!

src/listeners.js
ydidwania: `this.listeners` also contains the new `listeners` created during reconciling the `Hooks` Azure table with the `Queues` Azure table. So instead of stopping all the listeners in `this.listeners`, we only have to stop the `listeners`( aka `consumers`) of the `queues`( `queueName`=`$(hookGroupId)/$(hookId)` ) which are no longer in the `Hooks` Azure table.

A queue is pulled from the `Queues` list when its corresponding `hook` is found in the `Hooks` table.  So at the end of the `do..while` loop the `Queues` list has all the queues which are no longer of use to us, and must be deleted. and the corresponding listener stopped

src/listeners.js
ydidwania: Yes, `createListener` can do this, but we are not adding an entry to the `Queues` table everytime we are creating a new listeners. For example, when reconciling after a restart, we are only reclaiming the queues present in the `Queues` table by creating the listeners

ydidwania:
Yes, I forgot to include that. I think reconciling the `Hooks` table with the `Queues` and creating listeners during startup would be sufficient.

ydidwania:
@djmitche some missing scopes (maybe for the new Queues Azure table) are causing tests to fail. Also I had missed updating the bindings in the `Queues` Azure table when binding and unbinding exchanges from the queue

src/listeners.js
ydidwania: I found  `_.intersectionWith(oldBindings, newBindings, _.isEqual)` and used its diff with old and new bindings to unbind and bind respectively.

src/listeners.js
ydidwania: Yes, I did that in the last couple commits. Having the `while` loop also didn't guarantee completion of the loop before the next segment (where unused queues are deleted)  is executed.

djmitche:
Scopes are fixed.  I'll close and re-open to test that out.  It looks like this needs to get rebased over @Biboswan's changes, too.  Otherwise, is this ready for me to have another look?

djmitche:
Yes, go ahead.  I'll look now (I said I'd close/re-open and meant to check back, but I forgot to do both of those things!)

djmitche:
32617a09bebef8c1607742f4919ac45d3275fab0 fixes the tests :)

ydidwania:
Haha, I thought of setting `hook.bindings` as the default, but I hadn't saved the file :)

ydidwania:
The other test passes locally for me.
Also `'eslint-config-taskcluster'` doesn't allow me to use decorators

djmitche:
Regarding the decorator, the `@` syntax is really just a fancy way of writing a function call.  So
```js
@foo
function somefunc() {
}
```
is the same as
```js
somefunc = foo(function() {
});
```

So that might be a workaround for now, if it's not a simple matter of removing a lint check.  What's the lint error?

djmitche:
OK, what I wrote above might be incorrect :)

https://medium.com/google-developers/exploring-es7-decorators-76ecb65fb841

ydidwania:
I was trying to rebase, run into a bunch of errors.
The lint error was something like
```
Unexpected Token '@'
```

djmitche:
OK, that sounds like a more complex fix than just adjusting the .eslintrc.  So let's use the workaround I mentioned for now, and file a bug to update eslint.

ydidwania:
The workaround doesn't seem to work. When I write
```
this.reconcileConsumers = serialize(async () => {
 ... 
});
```
in the `setup()` method, all  calls to `reconcileConsumers` are  skipped and tests fail because of that.

djmitche:
I think I got the function signature for `serialize` wrong.  I think you'll need to call it like
```js
this.reconcileConsumers = serialize(this, 'reconcileConsumers', async () => {
});
```
does that work better?

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1514216 to try to allow actual decorators..

djmitche:
Hm, frustrating.  There's another approach to the same problem at
https://github.com/taskcluster/taskcluster-auth/blob/5077077f2e4094d22eef527dc9883cb8dffa387e/src/scoperesolver.js#L163-L169
which might be applicable here.  @imbstack is in the process of implementing a monorepo, after which time we can probably quickly get everything up to the same ES7 version and switch both hooks and auth to use `@serialize`.

ydidwania:
That worked  :)
Thats nice, some standardization will  surely help.
The only tests thats failing right now, seems to pass for me locally, is that also the case for you?

djmitche:
I'll check it out.

ydidwania:
Yes, those were extra changes that no longer had to be carried. I have removed all of those now. I also found a few trailing white spaces and removed them( there might be more).  

I factored out remove/have listeners into their own function in the latest commit which makes things a bit more readable, but you are right there is more that can be done. It is absolutely fine with me if you do those. Go ahead :)

ydidwania:
@djmitche I rebased all the commits and then I had to make some changes because of the version upgrade of `eslint` in [eslint-config-taskcluster](https://github.com/taskcluster/eslint-config-taskcluster). 
Also I was running into an error because of caching of injected components, so I changed the `withPulse` in `test/helper.js` to use the already injected `Taskcreator` component instead of loading it again.

djmitche:
Sounds great -- thanks for the notes on the rebase.

PR #110, opened 2018-09-28
Bug1430902: Format a invalid error message

alexandrasp:
Finally, I think we have a message with found errors delimited by `;`. Anyway,  I am not confident lol

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -204,10 +204,13 @@ builder.declare({
         '{{message}} in {{schedElement}}', {message: err.message, schedElement});
     }
   }
-
+  
   //handle an invalid schema
-  if (!ajv.validateSchema(hookDef.triggerSchema)) {
-    return res.reportError('InputError', 'Invalid Schema', {});
+  let valid = ajv.validateSchema(hookDef.triggerSchema);
+  if (!valid) {
+    return res.reportError('InputError', '{{message}}', {
+      message: ajv.errorsText(ajv.errors, {separator: '; '}),
+    });
   }
 
   // Try to create a Hook entity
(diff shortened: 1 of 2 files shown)


src/v1.js
```diff
@@ -204,10 +204,13 @@ builder.declare({
...
   }
-
+  
   //handle an invalid schema
-  if (!ajv.validateSchema(hookDef.triggerSchema)) {
-    return res.reportError('InputError', 'Invalid Schema', {});
+  let valid = ajv.validateSchema(hookDef.triggerSchema);
+  if (!valid) {
+    return res.reportError('InputError', '{{message}}', {
+      message: ajv.errorsText(ajv.errors, {separator: '; '}),
```
djmitche: It would be nice to format this as a markdown list, rather than a long line with semicolons, since the errors often line up nicely:
 * Property foo must be a string
 * Property bar must be a string
 * Property baz is required

djmitche:
Hi Alex!  Let me know when this is ready for review..

alexandrasp:
hey Dustin! I've just created a label, btw, is ready for review. I am not sure about the message format. I got ` ajv.errors dataPath and ajv.errors message`. Do you think in getting others fields from errors?

djmitche:
Looks good!  I added some whitespace in the tests, and also added a line to the error message so users know what field was wrong:

```
triggerSchema is not a valid JSON schema:
 * Property .properties['location'].type should be equal to one of the allowed values
 * Property .properties['location'].type should be array
 * Property .properties['location'].type should match some schema in anyOf
 * Property .type should be equal to one of the allowed values
 * Property .type should be array
 * Property .type should match some schema in anyOf
```

PR #115, opened 2018-10-25
Bug 1491551 - Apend last trigger info to a new table

Biboswan:
The bug aims at resolving two issues:
1. Keep record of tasks created 
2. Eliminate the max modify attempts.

So I'm starting with 1. 
@djmitche


src/lastFire.js
```diff
@@ -0,0 +1,24 @@
+const Entity = require('azure-entities');
```
djmitche: This looks good, but let's define this entity in `src/data.js` rather than in a new file (that's the more typical approach for Taskcluster services)

src/scheduler.js
```diff
@@ -112,6 +113,18 @@ class Scheduler extends events.EventEmitter {
         taskId: hook.nextTaskId,
         time: new Date(),
       };
+
+      try {
+        await this.LastFire.create({
+          hookGroupId: hook.hookGroupId,
+          hookId: hook.hookId,
+          taskId: lastFire.taskId,
+          taskCreateTime: lastFire.time,
```
djmitche: we'll want to capture the result here, too.

src/scheduler.js
```diff
@@ -112,6 +113,18 @@ class Scheduler extends events.EventEmitter {
...
+        });
+      } catch (err) {
+        debug('Failed to append lastfire with err: %s', err);
```
djmitche: you can send this to `this.monitor.reportError(err)`, so it gets handled as an exception we can see.

src/scheduler.js
```diff
@@ -170,19 +183,13 @@ class Scheduler extends events.EventEmitter {
...
   Details:
-
     ${errJson}
-
   The service will try again to create the task on the next iteration.
-
   Thanks,
   TaskCluster Automation
-
   P.S. If you believe you have received this email in error, please hit reply to let us know.`,
```
djmitche: Any particular reason to remove these blank lines?

src/scheduler.js
```diff
@@ -112,6 +113,18 @@ class Scheduler extends events.EventEmitter {
         taskId: hook.nextTaskId,
         time: new Date(),
       };
+
+      try {
+        await this.LastFire.create({
```
djmitche: Note that `triggerHoookCommon` in `api.js` also modifies the lastFire property of the Hook entity.  It might make sense to abstract that out to a common function that can do both `hook.modify` and `lastFire.create`.  In fact, I'm not sure why the TaskCreator doesn't do it?

src/scheduler.js
Biboswan: Here I considered the table for storing only success results. So shall we consider storing both success and error.

djmitche: Yes, I think both are important when debugging.

src/scheduler.js
Biboswan: yes i think it's better then to append lastfire info in `taskcreator`

Biboswan: or pass a callback function to `taskcreator.fire`.

Biboswan:
@djmitche I was thinking to resolve no. issue of `MAX_MODIFY_ATTEMPS` We can set like if attemps -> MAX_MODIFY Use Entity.create({..}, true).

djmitche:
I think the way to fix #2 is to stop trying to modify the table so much.  Once we are putting that data elsewhere, we can remove the modify calls..

Biboswan:
`nextTaskId` and `nextScheduledDate` are also modified when running scheduler. Oh but I see the warning is due to `triggerHook` and not `scheduler`

djmitche:
Good point -- we'll need to make sure that `triggerHook` does not modify `nextTaskId`

Biboswan:
Shall this be the approach for expiration process https://github.com/taskcluster/taskcluster-index/blob/master/src/data.js#L143. Instead of guessing I would like to know how is the `expires ` method in index called since it seems to be unused in that repo

djmitche:
> auth:azure-table:read-write:jungle/LastFireTestTable

Fixed.

> Instead of guessing I would like to know how is the expires method in index called since it seems to be unused in that repo

The expires method is called indirectly as a "[cron task](https://github.com/taskcluster/taskcluster-terraform/blob/master/index-service.tf#L61)". The `expire` string is matched against definitions in`Procfile` in the root directory.

Expiration for last-fire data will be a little more complicated, since we want to expire rows after a certain number of items per hook have been created, rather than just deleting all rows older than a specific time.

Note, too, that we should think about deleting rows from the last-fire table when deleting a hook.

Both expiration and deletion can wait for another PR (but that's something else you could get a start on now).  I'll have a look at this one soon.

src/scheduler.js
djmitche: I'd like to see the blank lines in this message restored.

src/taskcreator.js
```diff
@@ -79,7 +92,34 @@ class TaskCreator {
...
+    try {
+      await appendLastFire(hook, context, lastFire);
+    } catch (err) {
+      debug('Failed to append lastfire with err: %s', err);
+      this.monitor.reportError(err);
+    }
+
+    if (lastFire.error) {
+      return Promise.reject(lastFire.error);
+    }
```
djmitche: I don't think this is necessary -- the error has already been recorded in the LastFire table *and* reported via the monitor, so we can just return success here.

src/scheduler.js
Biboswan: oh sorry just missed it. Dont know why even i did this.

src/taskcreator.js
Biboswan: If there is an error notification may be send via email. https://github.com/taskcluster/taskcluster-hooks/blob/master/src/scheduler.js#L125

djmitche: Ah, good point.

djmitche:
This looks good!  Just a few issues with the tests to clean up.

I'll be away until Tuesday, but @owlishDeveloper might be able to help deploying once the tests pass.  Note that https://tools.taskcluster.net/auth/clients/project%2Ftaskcluster%2Ftc-hooks%2Fproduction will need a scope added for the new table.

djmitche:
green tests :D

owlishDeveloper:
@imbstack - I don't have deploy access to hooks. Halp :) I already added the scopes to the client.

PR #118, opened 2018-11-14
Bug1491551 contd

Biboswan:


src/taskcreator.js
```diff
@@ -46,6 +48,18 @@ class TaskCreator {
     return task;
   }
 
+  async appendLastFire({hookGroupId, hookId}, {firedBy}, lastFire) {
```
djmitche: It might be good to simplify this calling convention to just `{hookGropuId, hookId, taskId, taskCreateTime, firedBy, result, error}` and pull those properties from the various required places in the `fire` method.

src/taskcreator.js
```diff
@@ -79,7 +93,34 @@ class TaskCreator {
...
       return {status: {taskId: options.taskId}};
     }
-    return await queue.createTask(options.taskId, task);
+
+    let lastFire, taskCreateRes;
+    try {
+      taskCreateRes = await queue.createTask(options.taskId, task);
+      lastFire = {
+        result: 'success',
+        taskId: hook.nextTaskId,
```
djmitche: This is different from the taskId that was just created!  I think you want `options.taskId` here.  You'll also need the taskId even in the failure case (since taskId will be part of the rowKey).

src/taskcreator.js
```diff
@@ -79,7 +93,34 @@ class TaskCreator {
       this.lastCreateTask = {taskId: options.taskId, task};
       return {status: {taskId: options.taskId}};
     }
```
djmitche: Because of this `fakeCreate` conditional, it's hard to test the code that calls `appendLastFire`.  I think you were telling me that earlier, but I didn't understand :)

An alternative is to factor out the code that calls `queue.createTask` (basically the try/catch block below) into a separate method.  Then, in testing, just override that method on the `TaskCreator` instance with a dummy version.

Having done that, you can test both the "success" case (where the task-creation method returns something like `{result: 'success', taskId, time}`) and the "failure" case (`{result: 'error', taskId, error, time}`).

test/taskcreator_test.js
```diff
@@ -207,6 +208,21 @@ suite('taskcreator_test.js', function() {
           (err) => { debug('Got expected error: %s', err); });
       });
     }
+
+    test('adds a new row to lastFire', async function() {
```
djmitche: Let's do *two* appends for different hookIds, and see if they can be fetched independently.

src/data.js
```diff
@@ -177,3 +177,29 @@ Hook.prototype.definition = function() {
 
 // export Hook
 exports.Hook = Hook;
+
+const LastFire = Entity.configure({
+  version:              1,
+  partitionKey:         Entity.keys.StringKey('hookGroupId'),
+  rowKey:               Entity.keys.StringKey('taskCreateTime'),
```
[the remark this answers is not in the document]
Biboswan: I also preferred taskId but `Operation is not support for this data type`

src/data.js
```diff
@@ -177,3 +177,29 @@ Hook.prototype.definition = function() {
...
+    taskCreateTime:     Entity.types.Date,
+    result:             Entity.types.String,
+    error:              Entity.types.String,
+  },
+});
+
+LastFire.prototype.definition = function() {
+  return Promise.resolve({
+    hookGroupId:  this.hookGroupId,
+    hookId: this.hook,
```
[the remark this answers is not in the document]
Biboswan: i totally messed this bug since the beginning :(

src/data.js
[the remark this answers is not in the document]
djmitche: Hm, can you show the whole error?

test/taskcreator_test.js
Biboswan: keeping this test and adding a new test as well.

src/data.js
```diff
@@ -177,3 +177,29 @@ Hook.prototype.definition = function() {
...
+    error:              Entity.types.String,
+  },
+});
+
+LastFire.prototype.definition = function() {
+  return Promise.resolve({
+    hookGroupId:  this.hookGroupId,
+    hookId: this.hookId,
+  });
+};
```
djmitche: Is this function ever called?  If not, it can probably be deleted (it doesn't return very much data about the LastFire row!)

src/taskcreator.js
```diff
@@ -79,7 +93,42 @@ class TaskCreator {
...
+
+    try {
+      await this.appendLastFire({
+        hookGroupId: hook.hookGroupId,
+        taskCreateTime: lastFire.time, 
+        hookId: hook.hookId,
+        firedBy: context.firedBy,
+        taskId: lastFire.taskId,
+        result: lastFire.result,
+        error: lastFire.error || '',
```
djmitche: I think this needs to convert `lastFire.error` into a string (so `lastFire.error.toString()`.  Is there a test for the error case?

test/taskcreator_test.js
```diff
@@ -81,6 +82,51 @@ suite('taskcreator_test.js', function() {
       },
     };
 
+    const defaultHook2 = {
```
djmitche: Could this definition be shortened?  Something like `const defaultHook2 = _.assign(defaultHook, hookId: 'tc-test-hook2')`..

test/taskcreator_test.js
```diff
@@ -207,6 +253,72 @@ suite('taskcreator_test.js', function() {
...
+        hookGroupId: hook.hookGroupId,
+        hookId: hook.hookId,
+        taskId:  hook.nextTaskId,
+      });
+      assume(res.taskId).equals(hook.nextTaskId);
+    });
+
+    test('Fetch two appended lastFire rows independently', async function() {
+      let hook = _.cloneDeep(defaultHook);
+      let hook2 = _.cloneDeep(defaultHook2);
```
djmitche: In fact, maybe we don't even need a variable named `defaultHook2` -- the `_.assign(..)` I mentioned above could appear directly here.

PR #123, opened 2018-12-13
Bug 1512512 - remove listener each time it's added

djmitche:
I'm not sure if it's better to add/remove the listener for every test, or to just add it in suiteSetup and remove it in suiteTeardown..


owlishDeveloper:
Listener is just a function? I wonder if we need to add/remove it at all?

djmitche:
Good point - it could be added once in suiteSetup and removed once in suiteTeardown.  But it's easier to just click "Merge" now :)

PR #127, opened 2018-12-25
Bug 1514387 - use unique table names in testing to avoid conflicts

djmitche:


Biboswan:
All good. Just want to ask whether such conflict can occur in mock mode as well?

djmitche:
Good question -- no, in mock mode everything is in the process's memory, so if two tests are running in different processes they do not interfere with each other.

PR #130, opened 2019-01-03
hook.lastFire.error must be an object, not a string

djmitche:
/cc @Biboswan


djmitche:
I'm going to just land this as a hotfix.

